PR #16, opened 2015-02-27
Support for FOO-LICENSE and LICENSE-FOO filenames

### LINKED ISSUE #15: Detect MIT-LICENSE file
I don't know where the convention comes from, but I've seen enough repos with a `MIT-LICENSE` file that I'm guessing it was a recommended approach at some point. For example: https://github.com/github/brakeman.

Would it make sense to add support for this?

benbalter:
This adds support for repos with licenses in the `FOO-LICENSE` format and Fixes #15.

To avoid false positives, the default logic remains to look to the set list of known license files. If it doesn't find one, it'll then look for the first file in the project root that matches `/license/i`. 

This should allow for the literally millions of repos with license files names `LICENSE-MIT`, `MIT-LICENSE`, `GPL-LICENSE`, etc., although I'm not sure where that pattern came from.


nunobrito:
Good catch. This reminds the possibility of using a regular expression to catch cases where the LICENSE file has an extension (normally .md or .txt).

What do you think of something like?
```
(\b[\w]+-|)LICEN(C|S)E(.|)+
```

Should catch these combinations:
- LICENSE
- MIT-LICENSE
- LICENSE-MIT
- LICENCE-APACHE
- APACHE-LICENSE
- MIT-LICENSE.txt
- MIT-LICENSE.md
- LICENSE-MIT.txt

I've tested this on RegExr from http://www.regexr.com/

benbalter:
@nunobrito I like your thinking. Right now the logic is (1) is there a root-level file that directly matches one on the list? (2) Is their a root-level file that matches `/license/i`. The 2nd should cover the examples above, but wanted to prefer the whitelist to prevent a false positive from bumping an actual license. 

The `/(c|s)/` catch is a good catch.

nunobrito:
Ok, that makes good sense.

If it serves of any use, I've found some 40k LICENSE files on Github and about 300 LICENCE files. 

Probably not relevant but on GPL projects one might stumble on COPYING files instead of license. From what I gather, seems to be a file generated automatically. Notable example is the Linux kernel: https://github.com/torvalds/linux

However, these COPYING files are rare to come across. Only found 170 across the repositories.

PR #24, opened 2015-03-09
Add copyright to list of filenames

jontro:
COPYRIGHT is also used as specifying the license. 
Searching github gives 276499 matches for files named copyright.

### DIFF
--- a/lib/licensee/project.rb
+++ b/lib/licensee/project.rb
@@ -10,6 +10,7 @@ class Licensee
       license.md
       unlicense
       copying
+      copyright
     ]
 
     # Initializes a new project


benbalter:
@jontro interesting. Is it your sense that those files actually have licenses in them? If so, isn't it a bit of a misnomer? From my experience, they usually have explicit copyrights, no?

jontro:
@benbalter  I had the wrong search query. Searching "filename:copyright path:/" shows around 2 700 projects. I would assume most of the projects contain the licence, just like COPYING would. The reason I looked it up is because one of my projects had one (a quite old one). What do you mean about projects having explicit copyrights?

benbalter:
[here's the search](https://github.com/search?utf8=%E2%9C%93&q=copyright+filename%3Acopyright+path%3A%2F&type=Code&ref=searchresults) if anyone else would like to reproduce it.

> What do you mean about projects having explicit copyrights?

Like [this](https://github.com/caruccio/openshift-origin-cartridge-opentracker/blob/7ebd67b573a0ed32fa9b3ff8fd75951357b21779/COPYRIGHT), e.g., not a license, but a retention of copyright.

Since it's at the end of the list, I think it should be fine. The only edgecase is when a repository has a `COPYRIGHT` file with `(c) 2015 Ben Balter` and a `license-mit.txt` file, it'll prefer the copyright over the license, but think that's a pretty odd edgecase.

Thanks @jontro!

PR #26, opened 2015-03-10
Use also other filename than LICENSE in README

hroncok:
I had to inspect the source to see if only LICENSE files with different extensions are inspected, or other common files, such as COPYING are inspected as well. This makes it more clear.


benbalter:
Nice catch. Thanks @hroncok!

PR #32, opened 2015-06-24
Enable passing directory explicitly in `licensee` CLI tool

mislav:
This is so you don't have to change directory to look up a license


benbalter:
Makes sense. Thanks for the pull request! :tada:

PR #33, opened 2015-06-24
Handle non-git repos in Licensee::Project

### LINKED ISSUE #8: Support for non-git versioned repos
Potentially via FakeBlob if init fails for the project?

mislav:
I first tried to use Licensee::Project with a regular directory which wasn't a git repo. It failed with a Rugged exception and only by looking at the source I realized that Project could only read git repositories, not normal filesystem repositories.

This adds the feature of reading normal repositories if Rugged detects that the directory is not a git repo.

Fixes #8


benbalter:
Wow. This is awesome. Thanks for taking the time to bundle this together (and in a way I couldn't do). :octopus:

PR #35, opened 2015-06-29
Better file matching regex

benbalter:
Which exposes ~~`Licensee::Project.license_file?` and `Licensee::Project.maybe_license_file?` methods~~ `Licensee::Project.match_license_file` for use outside of Licensee.

//cc @mislav

### DIFF
--- a/lib/licensee/project.rb
+++ b/lib/licensee/project.rb
@@ -2,17 +2,6 @@ class Licensee
   class Project
     attr_reader :repository
 
-    # Array of file names to look for potential license files, in order
-    # Filenames should be lower case as candidates are downcased before comparison
-    LICENSE_FILENAMES = %w[
-      license
-      license.txt
-      license.md
-      unlicense
-      copying
-      copyright
-    ]
-
     # Initializes a new project
     #
     # path_or_repo path to git repo or Rugged::Repository instance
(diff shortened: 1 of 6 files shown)


lib/licensee/project.rb
```diff
@@ -45,6 +34,19 @@ def license
...
 
+    # Scores a given file as a potential license
+    #
+    # filename - (string) the name of the file to score
+    #
+    # Returns 1 if the file is definately a license file
+    # Return 0.5 if the file is likely a license file
+    # Returns 0 if the file is definately not a license file
+    def self.match_license_file(filename)
+      return 1   if self.license_file?(filename)
```
mislav: Should the match score be higher if the filename matches a file closer to the beginning of LICENSE_FILENAMES array? It seems that the names in that array are sorted by preference.

lib/licensee/project.rb
```diff
@@ -2,17 +2,6 @@ class Licensee
   class Project
     attr_reader :repository
 
-    # Array of file names to look for potential license files, in order
-    # Filenames should be lower case as candidates are downcased before comparison
-    LICENSE_FILENAMES = %w[
```
mislav: The backwards-incompatible consequence of this change is that the license filenames are not picked by the order of preference anymore. Now, the "COPYING" file will win over a "LICENSE" file in the same repo, when previously the opposite was the case.

benbalter: This is the question I asked in chat... if regex was read left to right. If so, then I believe LICENSE should win over COPYING which should win over COPYRIGHT.

mislav: You did and we talked about it, but I believe we reached a conclusion that it doesn't matter how regexes are read, because files are tested one by one in sequence. "COPYING" will match the regex and get assigned a score of "1", then LICENSE will match and also get assigned a score of "1", and so on.

benbalter: You're 100% right. Opened https://github.com/benbalter/licensee/pull/36 with a failing test.

PR #36, opened 2015-07-20
License preference

benbalter:
Given `LICENSE` and `COPYING.md` in the same repo, Licensee should prefer `LICENSE` over `COPYING` to preserve the old preference behavior.

For now, this PR adds a failing test by adding a second license to the `licenses.git` fixture.

/cc @mislav, https://github.com/benbalter/licensee/pull/35#discussion_r35033475


mislav:
Thanks. I think there was nothing wrong with the LICENSE_FILENAMES array as a record of which files we prefer, in order. I vote to bring it back, since it's tricky implement this kind of priority using regex.

benbalter:
@mislav Agree with you. What do you think about the approach in e796c9b?

Two motivations for doing things that way:

1. The match should be extension agnostic. Theoretically we could strip the extension prior to checking if it was in the array instead. My logic is that if a repo has a `LICENSE.txt` and `LICENSE` file, it's a toss up which one even a human would decide to look at.

2. I don't see an elegant way to score within the array. `1 - index / length`? I felt like the tiered regex was a more elegant approach and made the intention clear in terms of prioritizing license matches.

mislav:
I think your approach is good for now :+1:. It's clear in code and handles extensions gracefully. Yeah, scoring based on array index is definitely tricky and this is cleaner.

PR #37, opened 2015-07-20
Use yaml.safe_load

benbalter:
/cc @ptoomey3

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -30,7 +30,7 @@ class Licensee
 
     # License metadata from YAML front matter
     def meta
-      @meta ||= YAML.load(parts[1]) if parts && parts[1]
+      @meta ||= YAML.safe_load(parts[1]) if parts && parts[1]
     end
 
     # Returns the human-readable license name


lib/licensee/license.rb
```diff
@@ -30,7 +30,7 @@ def content
 
     # License metadata from YAML front matter
     def meta
-      @meta ||= YAML.load(parts[1]) if parts && parts[1]
+      @meta ||= YAML.safe_load(parts[1]) if parts && parts[1]
```
ptoomey3: Just because you're paranoid doesn't mean they aren't after you.

mislav:
Does this drop support for older Ruby versions? Do we care?

benbalter:
Believe it's Ruby >= 2.1. I'm fine with that, but I'll defer to y'all, as y'all are much better Rubyists than I. 

Alternatively we could check if the method exists and fall back to `Yaml.load`.

mislav:
Good point re: major bump.

I feel that it's worth checking whether the method exists. It's a bummer that this detail prevents Ruby 2.0 from using this library.

ptoomey3:
> I feel that it's worth checking whether the method exists. It's a bummer that this detail prevents Ruby 2.0 (preinstalled on OS X) from using this library.

This seems super reasonable given that the safe yaml use is purely a defensive paranoia thing to prevent any misuse/bugs in the future. As written now, there appears to be no risk.

PR #38, opened 2015-07-20
Check if safe_yaml exists before calling

benbalter:
Via https://github.com/benbalter/licensee/pull/37#issuecomment-123057842, the idea is that `safe_yaml` is a nice to have paranoia, but isn't strictly necessary. Rather than require a major bump/breaking change, lets use safe_yaml if we got it, and if not, fall back to vanilla `YAML.load`. 

This also adds a travis check for Ruby 2.0.

/cc @mislav, @ptoomey3, https://github.com/benbalter/licensee/pull/37

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -30,7 +30,13 @@ class Licensee
 
     # License metadata from YAML front matter
     def meta
-      @meta ||= YAML.safe_load(parts[1]) if parts && parts[1]
+      @meta ||= if parts && parts[1]
+        if YAML.respond_to? :safe_yaml
+          YAML.safe_load(parts[1])
+        else
+          YAML.load(parts[1])
+        end
+      end
     end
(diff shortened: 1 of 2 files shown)


mislav:
:+1:

lib/licensee/license.rb
```diff
@@ -30,7 +30,13 @@ def content
 
     # License metadata from YAML front matter
     def meta
-      @meta ||= YAML.safe_load(parts[1]) if parts && parts[1]
+      @meta ||= if parts && parts[1]
+        if YAML.respond_to? :safe_yaml
```
ptoomey3: Should this be `safe_load`?

benbalter: Ugh. Yes.

benbalter: Fixed via 36654dd.

PR #48, opened 2015-08-26
Add support for package managers

benbalter:
Disabled by default, but would detect a NPM/Bower/Gemspec license, if it's an exact match to an SPDX id.

Todo:

- [x] Moar tests
- [x] Add Rubygem support without parsing the Rubygem
- [x] ~~Confirm parsing arbitrary JSON is safe~~
- [x] Rename File to ProjectFile
- [x] Don't parse JSON, use Regex

/cc @bkeepers, @arfon for recent PRs where this would have helped.


wking:
@benbalter, can you provide motivation for [setting the package confidence to 90][1] (which you did in b09831ac9 after initially having 80 in 7f155326)?  It's [leading to `LICENSE.spdx` results getting 90% confidence][2], which I think is surprisingly low for [spdx-tutorial][3].  Is the thought “folks who wrote these package files have a 10% chance of getting their own project's license expression wrong”?  That wouldn't surprise me in general, although I'm 100% confident that @david-a-wheeler got his license expression right for spdx-tutorial ;).  And in the absence of conflicting information, I'd be inclined to claim 100% confidence in `LICENSE.spdx` values for “this is what the package maintainers are telling us is the license for the content of this package” (which is distinct from “this is the license for the content of this package”).

[1]: https://github.com/benbalter/licensee/pull/48/files#diff-752548ddc4704924a2b06c481a4831c9R9
[2]: https://github.com/benbalter/licensee/pull/241#discussion_r153303460
[3]: https://github.com/david-a-wheeler/spdx-tutorial/commit/99c84fd074564a991b79169e7e6eeb07f22d6475

mlinksva:
I don't know @benbalter's reasoning for a specific number, but here are my thoughts...

I think you have a point, but your confidence in a particular person is beside it. Licensee reports on confidence a given license ~notice matches a particular license. So if it finds something in a `LICENSE` file, and in `README`, and in a package manager file, it says how confident it is it found a particular license in each of those locations. It probably should report 100%  if it finds a single SPDX ID in the license field of a package manager file; that's an unambiguous reference to that license, just as a verbatim copy of a license in a `LICENSE` file is unambiguously that license.

Confidence that a license applies to a repo should be thought of as a different concept -- for which 100 (or even 90) given any single method is too high. I'd have the overall reported confidence go up with each corroboration, but never reach 100 (there could be conflicting info in places we haven't looked, people could lie or have no idea what they're doing, and licensee doesn't know anything about the reputation of any particular user). Of course the contribution of a particular method would be a matter for debate.

wking:
On Tue, Nov 28, 2017 at 09:26:38PM +0000, Mike Linksvayer wrote:
> It probably should report 100% if it finds a single SPDX ID in the
> license field of a package manager file; that's an unambiguous
> reference to that license, just as a verbatim copy of a license in a
> `LICENSE` file is unambiguously that license.

So you agree that the confidence reported for LICENSE.spdx should be
100%, not its current 90% [1]?

> Confidence that a license applies to a repo should be thought of as
> a different concept…

Agreed.

> …for which 100 (or even 90) given any single method is too high…

I'm less sure about this.

> … there could be conflicting info in places we haven't looked…

Agreed, but putting a number on this is going to be hard.  I think the
Licensee confidences should come after that (“from the places we've
looked, we're 90% sure the package license is …”).  If a repository
has a LICENSE.spdx declaring the package version and no conflicting
information in places where Licensee looked, I think it should be
reporting 100% confidence for both the file and the package as a
whole.

> … people could lie or have no idea what they're doing, and licensee
> doesn't know anything about the reputation of any particular user…

And while you could address this with something like PGP's web of
trust [2] and/or a plagarism-detector, that would be a lot of work ;).
I'm fine lumping this into the “things Licensee does not consider”
category, and having Licensee claim 100% confidence anyway.  We'd want
a big warning in the README, and updates to [3] to enumerate these
sorts of things.

[1]: https://github.com/benbalter/licensee/pull/241#issue-276239902
[2]: https://en.wikipedia.org/wiki/Pretty_Good_Privacy#Web_of_trust
[3]: https://github.com/benbalter/licensee/blob/v9.6.0/docs/what-we-look-at.md

PR #51, opened 2015-09-01
confirm license contains the license name or nickname if it should

benbalter:
Fixes https://github.com/benbalter/licensee/issues/50.

```
script/git-repo https://github.com/evan/memcached
Unknown
```

(We don't vendor the AFL)

As an added bonus, it actually speed things up a bit by letting us bail early.

The risk here, is a heavily modified license that would require us to go as far down the Matcher list as the Levenshtein matcher, may not have the license title (e.g., in the case of MIT) in which case we'd raise a false negative, but would like to try it in the wild a bit first to see how much that impacts things.

/cc @bkeepers

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -76,6 +76,14 @@ class Licensee
       meta.nil? ? key.capitalize : meta["title"]
     end
 
+    def nickname
+      meta["nickname"] if meta
+    end
+
+    def name_without_version
+      /(.+?)(( v?\d\.\d)|$)/.match(name)[1]
+    end
+
     def featured?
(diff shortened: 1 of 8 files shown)


bkeepers:
Clever.  :+1:  if it works!

benbalter:
In hindsight, could have also just added the AFL, but this actually has a significant performance improvement because it further limits the potential licenses levenshtein (our most expensive operation) needs to check.

benbalter:
> may not have the license title (e.g., in the case of MIT) in which case we'd raise a false negative

Added an explicit test to confirm the GitMatcher will still catch MIT, even if it's missing the title.

PR #52, opened 2015-09-01
Attribution detection and removal

benbalter:
Todo:

- [x] Tests
- [x] ~~Multiple copyright lines~~

Eventually:

Maybe build it from the package manager's author file

Fixes https://github.com/benbalter/licensee/issues/19

/cc @bkeepers

### DIFF
--- a/bin/licensee
+++ b/bin/licensee
@@ -10,6 +10,7 @@ license = project.matched_file
 if license
   puts "Matched file: #{license.path}"
   puts "License: #{license.match ? license.match.meta['title'] : 'no license'}"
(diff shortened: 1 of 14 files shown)


benbalter:
By stripping the attribution from the license before comparison, we actually speed things up a bit, because we're more likely to use exact matcher (and thus be more confident). Here's an example:

```
script/git-repo https://github.com/jekyll/jekyll
Matched file: LICENSE
License: MIT License
Attribution: Copyright (c) 2008-2015 Tom Preston-Werner
Confidence: 100%
Method: Licensee::ExactMatcher
```

bkeepers:
:metal:  this is great!

PR #55, opened 2015-09-03
Memoize false and nil values

benbalter:
/cc @spraints because Ruby. :smile:

### DIFF
--- a/lib/licensee.rb
+++ b/lib/licensee.rb
@@ -60,7 +60,7 @@ class Licensee
     end
 
     def package_manager_files?
-      @package_manager_files ||= false
+      @package_manager_files || false
     end
   end
 end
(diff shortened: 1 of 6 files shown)


spraints:
:+1: Looks good!

lib/licensee.rb
```diff
@@ -60,7 +60,7 @@ def confidence_threshold
     end
 
     def package_manager_files?
-      @package_manager_files ||= false
+      @package_manager_files || false
```
bkeepers: Does `@package_manager_files` ever get set?

benbalter: Package manager files is set via `Licensee.package_manager_files = true`. If the instance variable is true, it returns the instance variable. If the instance variable is `false`, it'd fall through the `||=` (as I learned yesterday) and would be `false` either way.

benbalter: Specifically `attr_writer :confidence_threshold, :package_manager_files`

bkeepers: Ah, thanks

Issue #50, opened 2015-09-01
Mis-detection of AFL 3.0 as OSL 3.0


bkeepers:
```
./script/git-repo https://github.com/evan/memcached                     
Matched file: LICENSE
License: Open Software License 3.0
Confidence: 98.28526890101325%
Method: Licensee::LevenshteinMatcher
```

https://github.com/evan/memcached/blob/master/LICENSE

benbalter:
Huh. Looks like the math is technically accurate. They are 98%+ similar. May need to add a sanity check that the license name appears in the file.

PR #61, opened 2015-09-24
Better Git interface + Dice classificator

vmg:
@benbalter @bkeepers: Here's what I've come up with. Full changelog ahead:

- Changed the way `ProjectFile` works. It has been specialized into `Project::LicenseFile` and `Project::PackageInfo`. We use more efficient Rugged APIs to load the blob contents, and we no longer allocate or load a file object until we know for sure we'll attempt to match it

- Changed the way the matchers are picked for each file: instead of a global list of matchers, `LicenseFile` and `PackageInfo` provide the matchers they want to run.

- Changed `LevenshteinMatcher` to `DiceMatcher`. It uses Dice's coefficient on the wordsets of the licenses. This is the main change that I'd like to test (cc @bkeepers). If it's not working good enough, this can be trivially removed and swapped for something else

- Changed `normalize_content` to create a *normalized word set* instead of a normalized string. This is more efficient for both comparison, generation, and allows us to perform Dice's coefficient.

- Removed the `GitMatcher`: with the efficient Rugged APIs, using the hashsig info was very cumbersome. I believe that with `ExactMatcher` and falling back to `DiceMatcher` we should have the same accuracy

- Removed support for `Licensee::FileSystemRepository`: with Rugged specialized APIs for blob access, the filesystem repo abstraction was very leaky/hacky. It could be brought back, but I don't see us analyzing anything outside of a Git repo anytime soon, so I chose to remove the abstraction.

- Moved `Licensee.package_manager_files` to `Licensee::Project.new(path, detect_packages: true)`. Minor change, but it felt to me like keeping this as a global was a design mistake (and it was really showing on the unit tests -- I believe they look cleaner now).

In summary: the external API is mostly unchanged. I made sure that no features were removed or regressed. The library should work just the same, but it handles Git data much more smartly. I'm quite happy with the internal refactoring -- overall the design feels more elegant and it is measurably much more efficient re Git. Besides that, Levenshtein has been replaced with Dice: that's the thing I'm not sure about. I'd like your feedback and testing on that. Can you give it a go with some real life data?


bkeepers:
> - Changed `LevenshteinMatcher` to `DiceMatcher`. It uses Dice's coefficient on the wordsets of the licenses. This is the main change that I'd like to test (cc @bkeepers). If it's not working good enough, this can be trivially removed and swapped for something else

:+1: Investigating now.

> Removed support for Licensee::FileSystemRepository: with Rugged specialized APIs for blob access, the filesystem repo abstraction was very leaky/hacky. It could be brought back, but I don't see us analyzing anything outside of a Git repo anytime soon, so I chose to remove the abstraction.

We are using this to analyze package dependencies like rubygems and npm modules.

vmg:
> We are using this to analyze package dependencies like rubygems and npm modules.

Copy that!

vmg:
Now available: `Licensee::GitProject` and `Licensee::FSProject`, to analyze git repositories and folders on the FS, respectively. Also, I've removed a ton of duplicated code from `script/benchmark`. It seems to run now!

bkeepers:
Here are the changes from master for the top 100 repos on GitHub. These are all interesting cases.

```diff
--- master.txt	2015-09-24 15:48:39.000000000 -0400
+++ dice.txt	2015-09-24 15:48:49.000000000 -0400
@@ -10 +10 @@
-meteor/meteor: unknown
+meteor/meteor: MIT License
@@ -43 +43 @@
-harvesthq/chosen: unknown
+harvesthq/chosen: MIT License
@@ -55 +55 @@
-django/django: BSD 3-clause "New" or "Revised" License
+django/django: BSD 3-clause Clear License
@@ -71 +71 @@
-tastejs/todomvc: unknown
+tastejs/todomvc: MIT License
```

bkeepers:
I tested against a bunch of gems that are used on another project and here were the differences:

```
diff -U 0 master.txt dice.txt
--- master.txt	2015-09-24 16:08:02.000000000 -0400
+++ dice.txt	2015-09-24 16:02:41.000000000 -0400
@@ -87 +86 @@
-memcached: unknown
+memcached: osl-3.0
@@ -102 +101 @@
-net-ldap: unknown
+net-ldap: mit
```

memcache is actually AFL(https://github.com/evan/memcached/blob/master/LICENSE). See #50 for older versions of linguist mis-detecting this.

net-ldap is technically MIT, but there is enough [additional text in the license file](https://github.com/evan/memcached/blob/master/LICENSE) that I wonder if it should actually be detected as such.

vmg:
So, I had a chat with @bkeepers yesterday and we seem to agree the Dice classifier is in a pretty good spot. All those different classifications are actually *more* accurate than the previous implementation. Dice is detecting licenses that have some more extra text because it considers that text to be mostly redundant (it is in all cases). That's a good thing.

The only real issue is AFL vs OSL. As @benbalter pointed out in the older PR, he added the "language name in license" hack because it was making most calculations faster thanks to shortcutting. With the current speed of Dice, however, it feels the right approach would be to simply add AFL as a detected license so it just matches properly.

@bkeepers: What would be the steps to do that? Do I have to open a PR on choosealicense?

lib/licensee/matchers/dice_matcher.rb
```diff
@@ -0,0 +1,65 @@
...
+
+      private
+      # Calculate percent changed between file and potential license
+      def similarity(license)
+        overlap = (@file.wordset & license.wordset).size
+        total = @file.wordset.size + license.wordset.size
+        100.0 * (overlap * 2.0 / total)
+      end
+
+      def includes_license_name?(license)
```
benbalter: Are these still used anywhere? If not used internally, I'd say safe to :skull:

benbalter:
Wow. @vmg, once again, it's very clear that you put a lot of time and thought into this. This is awesome. Thank you.

Two quick questions that can be answered after this merges:

1. Unless I'm misreading, it looks like we lost tests for non-git repos? I don't know that it needs to have parity with git repo tests, but we should probably at least test that it works to prevent a regression.

2. If I understand correctly, a set is an unordered list without duplicates. Is it possible that a license is modified in such a way that the order is changed to modify the meaning, or that additional text is added, but using words already in the license? Based on our real-world tests, it seems like it's accurate enough, but I'm wondering if there's a strange edge case we're missing. Should we bump the confidence to say, 95?

> Changed the way ProjectFile works. It has been specialized into Project::LicenseFile and Project::PackageInfo. 

That was on my wish list for 6.x. Nice work.

> We use more efficient Rugged APIs to load the blob contents, and we no longer allocate or load a file object until we know for sure we'll attempt to match it

Nice.

> Changed the way the matchers are picked for each file

Wow. Hadn't thought of that, but now it's obvious. Nice.

> Moved Licensee.package_manager_files to Licensee::Project.new(path, detect_packages: true). Minor change, but it felt to me like keeping this as a global was a design mistake (and it was really showing on the unit tests -- I believe they look cleaner now).

Makes sense. 

> overall the design feels more elegant 

MUCH. Thank you.

> it feels the right approach would be to simply add AFL as a detected license so it just matches properly.

Agreed. I can take care of getting that added to choosealicense right now.

> So, I had a chat with @bkeepers yesterday and we seem to agree the Dice classifier is in a pretty good spot. 

Nice. Running through those repos was going to be my test case, and if @bkeepers is happy with the results, I'm happy.

Going to merge here, so we can test more easily, and will go get AFL added to CAL.

benbalter:
Also, not entirely scientific, but compare [2.3 seconds to run through all fixtures, modified and rewraped](https://travis-ci.org/benbalter/licensee/jobs/79330772), veruses [.071 today](https://travis-ci.org/benbalter/licensee/jobs/820243970)

bkeepers:
> Also, not entirely scientific, but compare 2.3 seconds to run through all fixtures, modified and rewraped, veruses .071 today

Yeah, new version is incredibly fast. I ran it against ~200 some gems in the blink of an eye.

vmg:
> Unless I'm misreading, it looks like we lost tests for non-git repos? I don't know that it needs to have parity with git repo tests, but we should probably at least test that it works to prevent a regression.

:sleepy: Somehow I screwed up that refactoring. :point_right: https://github.com/benbalter/licensee/pull/62

> If I understand correctly, a set is an unordered list without duplicates. Is it possible that a license is modified in such a way that the order is changed to modify the meaning, or that additional text is added, but using words already in the license? Based on our real-world tests, it seems like it's accurate enough, but I'm wondering if there's a strange edge case we're missing. Should we bump the confidence to say, 95?

I talked briefly with @bkeepers about this. Short answer is, yes, if you're malicious you can rearrange the text of a license to make it mean something else, and the classifier would still pick it up... Besides that corner case, I genuinely don't think that legitimate licenses will be miss-classified under any circumstances. The 90% rating we have right now sounds pretty good to me, because as you saw earlier it handles cases of licenses with some redundant information quite gracefully.

> Going to merge here, so we can test more easily, and will go get AFL added to CAL.

:heart_eyes: You even gave me push access to the repo! :kissing:

benbalter:
> You even gave me push access to the repo! 

Honestly, thought you had it from the last time around. Think line-for-line you've written most of the plugin at this point (at least the part that's not terrible). :smile_cat:

PR #63, opened 2015-09-25
WIP: Add AFL license

benbalter:
WIP, because we're having issues differentiating from OSL, which appears 98% similar:

```
  1) Failure:
TestLicenseeVendor#test_: different line lengths should detect the /Users/benbalter/projects/licensee/vendor/choosealicense.com/_licenses/afl-3.0.txt license.  [/Users/benbalter/projects/licensee/test/functions.rb:42]:
expeceted afl-3.0 but got osl-3.0 for .match. Confidence: 98.46827133479212. Method: Licensee::Matchers::Dice.
Expected: "afl-3.0"
  Actual: "osl-3.0"


  2) Failure:
TestLicenseeVendor#test_: different line lengths when modified should detect the /Users/benbalter/projects/licensee/vendor/choosealicense.com/_licenses/afl-3.0.txt license.  [/Users/benbalter/projects/licensee/test/functions.rb:42]:
expeceted afl-3.0 but got osl-3.0 for .match. Confidence: 98.46827133479212. Method: Licensee::Matchers::Dice.
Expected: "afl-3.0"
  Actual: "osl-3.0"
```

For different line lengths, the exact matcher should match. Going to investigate.

For the modified + different line lengths, given the speed, perhaps we can score all potential licenses and choose the best?

/cc @bkeepers, @vmg

### DIFF
--- a/lib/licensee/matchers/dice_matcher.rb
+++ b/lib/licensee/matchers/dice_matcher.rb
@@ -9,9 +9,7 @@ class Licensee
       # than the confidence threshold
       def match
         return @match if defined? @match
-        @match = potential_licenses.find do |license|
-          similarity(license) >= Licensee.confidence_threshold
-        end
+        @match = matches.empty? ? nil : matches.max_by { |l, sim| sim }[0] 
       end
 
       # Sort all licenses, in decending order, by difference in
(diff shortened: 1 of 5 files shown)


benbalter:
@vmg in 76e8c8a I modified the Dice matcher to score all potential licenses and return the best result (to account for OSL and AFL being so similar). Two questions:

1. Is this a bad idea? Given how cheap comparision now is, It didn't seem to degrade performance.

2. Is there a more elegant way to implement this? I added a `matches` method, but couldn't find a simpler way.

vmg:
> Is this a bad idea? Given how cheap comparision now is, It didn't seem to degrade performance.

It's not the best idea, but it seems it's the only way to catch these two extremely similar licenses. :joy:

> Is there a more elegant way to implement this? I added a matches method, but couldn't find a simpler way.

Your matches method was not being memoized, you were matching all licenses *twice*. I moved the logic inside the `match` getter. I think this is good to ship now. :)

PR #68, opened 2015-10-24
Add README detection

bkeepers:
There are a _lot_ of projects that will put the license in the README. While we can't reliably detect them all the time, I have observed that most of them add a heading called `## License` to their README. This is an attempt to spike out an implementation that will look for license content in the README.

```
$ script/git-repo https://github.com/juliangruber/isarray
License: MIT License
Confidence: 99.45945945945947%
Method: Licensee::Matchers::Dice
```

Thoughts?

@benbalter @vmg


benbalter:
Just to confirm, it's scans the README for a block that begins with `# License` to the next `#` that begins a new line (or the end of the file), pulls that block out, and then runs it through the normal comparison/identification pipeline?

From a using licensee to ensure our own compliance perspective, the lawyer in me says there's no difference between it being in the readme or in it's own file, as long as the text is there (versus a simple `License: GPL`.

From a using licensee to identify licenses on GitHub.com perspective, the open source developer in me says we have an opportunity to standardize how license information is distributed within an open source project, and what we support can go a long way to incentivize behavior one way or the other.

Assuming the above, as long as both package manager and readme detection is off by default (and thus likely off on GitHub.com), I'm :+1: and think it gets us the best of both worlds.

bkeepers:
> Just to confirm, it's scans the README for a block that begins with `# License` to the next `#` that begins a new line (or the end of the file), pulls that block out, and then runs it through the normal comparison/identification pipeline?

Yup. Works with any level of heading.

> I'm :+1: and think it gets us the best of both worlds.

:cool: I'll test this branch out a little more before merging.

vmg:
@bkeepers: Do we have data that shows that Readme detection will capture licenses in a significant number of repositories we're currently missing? I'm worried about bringing new code into the Ernicorns without a specific purpose.

The code itself looks correct but very limited in scope. We're only tracking MD documents that have a "License" header. Most of the time (really, all of the time -- as far as all the projects I've just casually checked), the content in the README will not be a full license file, but the "short license header" of the license, which we're not currently equipped to detect.

Are we sure this is the right approach?

benbalter:
> I'm worried about bringing new code into the Ernicorns without a specific purpose.

Based on my understanding from reading above, I believe @bkeepers use case here would only be to run the code for local detection (off by default, with no intention to turn it on on GitHub.com at this time)... we could even not include the file unless the option is toggled if we want to keep things slim.

bkeepers:
> Do we have data that shows that Readme detection will capture licenses in a significant number of repositories we're currently missing? I'm worried about bringing new code into the Ernicorns without a specific purpose.

I don't have very much data on it, but on the project I just tested this against, it found the licenses for about 10 of the 80 dependencies with undetected licenses (of about 400 total deps, so 2.5%). It's been the most common pattern of undetected licenses I've seen across Ruby and Node projects so far.

benbalter:
@bkeepers I'm hoping to cut a new minor release with https://github.com/benbalter/licensee/pull/70/files soon, and am glad to add this in if it's disabled by default (and with the intention that we're not going to be using it within GitHub.com any time in the foreseeable future) if it's helpful for using-in-the-wild purposes.

bkeepers:
@benbalter 👍  It's been really helpful in the cases I've tried.

@vmg what do you think?

hyandell:
Perhaps you could have this in as a hint, or a lower level of belief.
"License unidentified, README suggests it may be MIT".

benbalter:
@bkeepers still interested in this?

bkeepers:
> @bkeepers still interested in this?

Yep! This code works great for me. Are there any other changes you would want to see before this lands?

I like @hyandell's suggestion, but won't have time to make changes to the code in the next week or two.

benbalter:
>  Are there any other changes you would want to see before this lands?

Mainly just resolving the merge conflict.

> I like @hyandell's suggestion, but won't have time to make changes to the code in the next week or two.

That's what I hoped the confidence level could be for, so that someone who just wanted a best guess could set the confidence threshold accordingly.

bkeepers:
> Mainly just resolving the merge conflict.

Done!

lib/licensee/project.rb
```diff
@@ -17,7 +22,7 @@ def license
     end
 
     def matched_file
-      @matched_file ||= (license_file || package_file)
+      @matched_file ||= (license_file || readme || package_file)
```
benbalter: Readme before package file? What's your reasoning here?

benbalter: Oh, because it's the text, not the key. Carry on.

benbalter:
:sparkling_heart: :sparkles: :tada: 

Nice work. Merging, so that I can un-conflict https://github.com/benbalter/licensee/pull/77/files.

PR #75, opened 2015-11-19
More usage info for README

hyandell:
If useful - added more info on how to call licensee. Also moved the command line usage above the library usage as it gets a user trying out the library quicker


benbalter:
Thanks!

PR #76, opened 2015-11-26
Expose license hash

### LINKED ISSUE #67: Recognize non-standard licenses
Uniquely identifying non-standard licenses would make it possible to whitelist them or detect future changes in the text of the licenses.

Currently, repositories such as [github/linguist](https://github.com/github/linguist), which have many licenses in submodules to validate, must use some custom scripts or rely on manual validations. 
Having licensee generate a unique identifier for each non-standard license would make it possible:
- to track any future modification(s) in the license text;
- and to whitelist a few common licenses, such as the [TextMate bundles license](https://github.com/textmate/ant.tmbundle#license).

The unique identifier could eventually be a hash value of the license text.

benbalter:
This pull request exposes `License#hash` and `LicenseFile#hash` methods for identifying and comparing non-standard licenses.

This is implemented by adding `hash` and `content_normalized` methods to `ContentHelper`, which simply strip whitespace and calculates the SHA256 hash of the file's content. I'm not beholden to SHA256, if folks like @vmg have strong opinions for one hashing algorithm over another. Given the use case, it doesn't need to be cryptographic, but we'd want it to be fast and unique.

This isn't used for the built in license detection. The use case in mind here is the one described in #67. Where Licensee fails to detect a license, the hash could be used by end users to identify identical (once normalized, including removing copyright information) but unknown licenses.

I also moved `LicenseFile` and `ProjectInfo` into their own files, to make it a bit easier to test things more granularly, and standardized `License` and `LicenseFile` to use `#content` (rather than `License` using `#body` before) to describe the contents of the license (but aliased to `#body` to preserve backwards compatibility).

Fixes #67

### DIFF
--- a/lib/licensee.rb
+++ b/lib/licensee.rb
@@ -2,8 +2,13 @@ require_relative "licensee/version"
 require_relative "licensee/content_helper"
 require_relative "licensee/license"
 require_relative "licensee/project"
-require_relative "licensee/project_file"
 
+# Project files
+require_relative "licensee/project_file"
+require_relative "licensee/project_files/license_file.rb"
+require_relative "licensee/project_files/package_info.rb"
+
+# Matchers
 require_relative "licensee/matchers/exact_matcher"
 require_relative "licensee/matchers/copyright_matcher"
 require_relative "licensee/matchers/dice_matcher"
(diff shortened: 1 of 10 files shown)


pchaigno:
Awesome! :tada:

PR #78, opened 2016-01-04
WIP: Make rugged dependency optional

benbalter:
Per https://github.com/benbalter/licensee/issues/66 this makes the dependency on Rugged optional (falling back to the file system project method).

I'm sympathetic to the fact that needing to compile Rugged adds complexity to local use/development, but am concerned that this approach, although it works, adds significant complexity to the testing process.

On the one hand, we probably should run the test suite as a whole twice, once as a Git repo and once as a file system repo, but the way the tests are currently structured, and the fixtures, that's not the case.

@vmg, @bkeepers any strong opinions one way or the other? I'm fine punting on this for the time being.

/cc @sschuberth, fixes https://github.com/benbalter/licensee/issues/66

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,7 @@ rvm:
 script: "script/cibuild"
 sudo: false
 cache: bundler
+env:
+  matrix:
+    - RUGGED=0
+    - RUGGED=1
--- a/Gemfile
+++ b/Gemfile
@@ -1,3 +1,7 @@
 source "https://rubygems.org"
 
 gemspec
+
+if ENV["RUGGED"] == "1"
+  gem 'rugged', "~> 0.23"
+end
--- a/bin/licensee
+++ b/bin/licensee
@@ -3,7 +3,7 @@ require_relative "../lib/licensee"
 
 path = ARGV[0] || Dir.pwd
 
-project = Licensee::GitProject.new(path, detect_packages: true, detect_readme: true)
+project = Licensee.project(path, detect_packages: true, detect_readme: true)
 file = project.matched_file
 
 if project.license_file
--- a/lib/licensee.rb
+++ b/lib/licensee.rb
@@ -21,6 +21,12 @@ require_relative "licensee/matchers/package_matcher"
 require_relative "licensee/matchers/gemspec_matcher"
 require_relative "licensee/matchers/npm_bower_matcher"
 
+begin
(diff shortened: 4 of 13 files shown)


sschuberth:
Sorry if I'm missing something obvious and even getting slightly off topic, but what's the purpose of a `GitProject` anyway as we have a `FSProject` now? I mean, a Git working tree also is an `FSProject` as it's located on the file system. I would see the benefit of a separate `GitProject` if it was able to also scan remote Git repos without cloning them / checking them out, but that does not seem to be the case, or?

benbalter:
> I would see the benefit of a separate GitProject if it was able to also scan remote Git repos without cloning them / checking them out, but that does not seem to be the case, or?

@sschuberth that's exactly how Licensee is used within GitHub.com. Given the number of repositories (10M and counting), it's preferable to work in Git, rather than shuttle repository data back and forth.

sschuberth:
I see.

Another thought, more on-topic: Instead of making the distinction between `GitProject` and `FSProject` within the test methods, how about maintain completely separate tests instead, an expect the `GitProject` ones to fail if *rugged* is not installed?

sschuberth:
> [...] it's preferable to work in Git, rather than shuttle repository data back and forth.

@benbalter I was trying to (again) make sense out of what you said. I believe the advantage of `GitProject` over `FSProject` is not about not having to shuttle data back and forth. Even a `GitProject` requires you to have the Git repository on your local machine. It cannot magically scan a remote Git repository without cloning it. *However*, the clone may be a `--bare` one. Or if you are running `licensee` on the Git server itself, it can directly scan the bare repositories from there. So the advantage is that a `GitProject` does not require a working tree. I tried to make that clear as part of PR #91.

Would you agree to that I'm saying here?

benbalter:
@sschuberth See e.g., the context around GitRPC in http://githubengineering.com/move-fast/#preparing-for-an-experiment. We have the ability to run Git commands on remote machines, with only the result crossing the network. In our case, the files do not exist on disk, as we're hitting the Git DB directly for performane reasons.

sschuberth:
Ah, I didn't know about GitRPC, thanks for the link!

benbalter:
Closing as stale and too much a technical lift for me right now. If anyone else wants to take this on, glad to review a PR.

PR #84, opened 2016-02-29
s/http/https/

jdmaturen:


benbalter:
Ahh nice catch. Thanks @jdmaturen! :link: :lock:

PR #86, opened 2016-03-16
test bsd/isc/mit licenses before titles added/changed still detected

mlinksva:
If this is to be merged at all it probably shouldn't be until https://github.com/github/choosealicense.com/pull/373 is, which this is intended to test.

I [wondered](https://github.com/github/choosealicense.com/pull/371#issuecomment-196962274) whether adding/changing license titles as discussed in link above and the PR it builds on would cause any problems for license matching so additional licensee tests seemed like a good way to proceed.


benbalter:
It *should* be fine. It'll fall back to using the wordset comparison, rather than an exact match, which'll be slightly slower, but should still be a match, as the wordset should still be the same.

mlinksva:
It does still match, unless I wrote the tests wrong. I went ahead and merged the title adds/change in https://github.com/github/choosealicense.com/pull/373

benbalter:
Thanks @mlinksva!

PR #87, opened 2016-03-17
licensee <file> usage

mlinksva:
allow ad hoc cli matching of individual license files not in a
particular project location

I might be the only person in the universe who wants this, in which case I'm happy to create a ruby script for myself that [ab]uses licensee rather than have this be in licensee.

### DIFF
--- a/bin/licensee
+++ b/bin/licensee
@@ -6,9 +6,7 @@ path = ARGV[0] || Dir.pwd
 def print_file(license_file)
   if license_file
     puts "License file: #{license_file.filename}"
-    # if license_file.attribution
     puts "Attribution: #{license_file.attribution}" if license_file.attribution
-    # end
   end
 end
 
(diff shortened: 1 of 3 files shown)


bin/licensee
```diff
@@ -3,21 +3,34 @@ require_relative '../lib/licensee'
 
 path = ARGV[0] || Dir.pwd
 
-options = { detect_packages: true, detect_readme: true }
-project = Licensee::GitProject.new(path, options)
-file = project.matched_file
+def print_file(license_file)
+  if license_file
+    puts "License file: #{license_file.filename}"
+    # if license_file.attribution
```
benbalter: Looks like there's some stray debug comments in there

mlinksva: Oops, fixed.

benbalter:
Thanks @mlinksva!

PR #89, opened 2016-03-19
Fix warnings

pchaigno:
I had the following warning while running tests: `lib/licensee/project.rb:26: warning: shadowing outer local variable - name`. This pull request should fix it.


benbalter:
Thanks @pchaigno!

PR #90, opened 2016-04-14
Ensure files are read with UTF-8 encoding by default

sschuberth:
On Windows, simply running

    $ bin/licensee LICENSE.md

resulted in

    content_helper.rb:22: in `gsub!': incompatible encoding regexp match
    (UTF-8 regexp with IBM437 string)

because both the input file and the license files (like "afl-3.0.txt"
which contains Unicode quotes) were read as IBM437-encoded strings. Fix
this by explicitly reading them as UTF-8 strings.

### DIFF
--- a/bin/licensee
+++ b/bin/licensee
@@ -21,7 +21,7 @@ def print_evaluation(file)
 end
 
 if File.file?(path)
-  contents = File.open(path).read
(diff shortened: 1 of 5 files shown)


benbalter:
Makes sense. Thanks @sschuberth!

PR #93, opened 2016-05-03
Commonly use Licensee.project() both for scanning files and directories

sschuberth:
Defer the file check to the begin / rescue block in Licensee.project() as
creating a GitProject would fail for a file.


benbalter:
Makes sense. Thanks @sschuberth!

PR #94, opened 2016-05-03
Support running tests on Windows

sschuberth:
@benbalter This should be good to go now.

### DIFF
--- a/test/test_licensee_bin.rb
+++ b/test/test_licensee_bin.rb
@@ -3,8 +3,10 @@ require 'helper'
 class TestLicenseeBin < Minitest::Test
   should 'work via commandline' do
     root = File.expand_path '..', File.dirname(__FILE__)
-    Dir.chdir root
-    stdout, stderr, status = Open3.capture3("#{root}/bin/licensee")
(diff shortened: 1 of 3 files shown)


test/test_licensee_bin.rb
```diff
@@ -18,8 +19,9 @@ class TestLicenseeBin < Minitest::Test
 
   should 'work via commandline with file argument' do
     root = File.expand_path '..', File.dirname(__FILE__)
-    Dir.chdir root
-    stdout, stderr, status = Open3.capture3("#{root}/bin/licensee LICENSE.md")
+    stdout, stderr, status = Dir.chdir root do
+      Open3.capture3("#{windows? && 'ruby'} bin/licensee LICENSE.md")
```
benbalter: Think this may have been my original wrong doing, but we should be passing 2-3 arguments (ruby, bin, license), rather than one concatenated string, to allow Open3 to handle escaping and other command-line concerns (although less of a concern since these are tests).

benbalter:
Awesome. Thanks @sschuberth!

PR #106, opened 2016-08-01
Force encoding when reading in licenses

benbalter:


benbalter:
Not necessary after all.

PR #110, opened 2016-09-13
harmonize scoring based on extension (fixes #109)

### LINKED ISSUE #109: Score COPYING over COPYING.foo files
Would need to update the scoring in https://github.com/benbalter/licensee/blob/master/lib/licensee/project_files/license_file.rb#L18.

waldyrious:
This change implements the following ranking:

1. No extension: LICENSE, COPYING, etc.
3. Filetype extension (syntactic annotation): UNLICENSE.txt, COPYRIGHT.md, etc.
2. Content-related extension (semantic annotation): COPYING.image, LICENCE.go, etc.

### DIFF
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -14,10 +14,13 @@ module Licensee
 
       def self.name_score(filename)
         return 1.0 if filename =~ /\A(un)?licen[sc]e\z/i
-        return 0.9 if filename =~ /\A(un)?licen[sc]e\.(md|markdown|txt)\z/i
-        return 0.8 if filename =~ /\Acopy(ing|right)(\.[^.]+)?\z/i
-        return 0.7 if filename =~ /\A(un)?licen[sc]e\.[^.]+\z/i
-        return 0.5 if filename =~ /licen[sc]e/i
+        return 0.9 if filename =~ /\Acopy(ing|right)\z/i
+        return 0.8 if filename =~ /\A(un)?licen[sc]e\.(md|markdown|txt)\z/i
+        return 0.7 if filename =~ /\Acopy(ing|right)\.(md|markdown|txt)\z/i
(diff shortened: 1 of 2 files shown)


lib/licensee/project_files/license_file.rb
```diff
@@ -14,10 +14,13 @@ def attribution
 
       def self.name_score(filename)
         return 1.0 if filename =~ /\A(un)?licen[sc]e\z/i
-        return 0.9 if filename =~ /\A(un)?licen[sc]e\.(md|markdown|txt)\z/i
-        return 0.8 if filename =~ /\Acopy(ing|right)(\.[^.]+)?\z/i
-        return 0.7 if filename =~ /\A(un)?licen[sc]e\.[^.]+\z/i
-        return 0.5 if filename =~ /licen[sc]e/i
+        return 0.9 if filename =~ /\Acopy(ing|right)\z/i
```
benbalter: I think `LICENSE.md` should still take precedence over `COPYING` when both our present (but agree `COPYING` should take precedence over `LICENSE.image`)

benbalter:
@waldyrious This is great. One comment on the priorities, and then we'll just need to update the corresponding tests.

lib/licensee/project_files/license_file.rb
waldyrious: Sure, I'll change that.

waldyrious:
Shouldn't the tests be relative (score of form X > score of form Y) rather than binded to absolute values?

benbalter:
> Shouldn't the tests be relative (score of form X > score of form Y) rather than binded to absolute values?

If there's a concise way to write that, that would better match our intent here, yes, I believe the actual values can be treated as arbitrary.

waldyrious:
fixed the tests and rebased the branch.

benbalter:
Thanks @waldyrious! I'll get those rubocop errors clean up on `master` as I believe I may have introduced at least one of them. 😄

waldyrious:
Thanks :)

PR #112, opened 2016-09-13
Fix for concatenated licenses

### LINKED ISSUE #111: License is not properly detected with MIT licenses is appended with another license
Example:

1. Take MIT License
2. Append GPL license in the same file
3. Run licensee

Expected:

No license detected

Actual:

GPL license detected.

benbalter:
Fixes #111.

There's a few things going on here, that I can potentially cherry-pick into a separate PR if anyone feels strongly:

* Organize the tests into namespaced folders
* Add a test for the ContentHelper
* The exact matcher should also take into account length, not just similarity of words
* Test the PackageMatcher
* Make the CopyrightMatcher a bit more flexible, to support the ZLib license's copyright line, which beings with `(c)` not `copyright`
* Increase confidence threshold to 95%
* Based "max delta" on license length, not the file to be tested

@mlinksva I'd especially like your feedback on the changes here. Practically speaking, here's what I'm proposing, in terms of license detection behavior:

1. The confidence threshold has been increased from 90 to 95 percent similarity to avoid false positives. 

2. For a file to be an exact match, it must (A) be the same length, once normalize, and (B) contain only the same words. Previously it only looks at (B).

3. The CopyrightMatcher now strips lines that begin with `(C)` followed by a date, not just `Copyright`. This is necessary because the ZLib license doesn't have Copyright in the template.

4. The "max delta" is now calculated on the known license's length, not the file we're checking. Specifically, if you have a known license that's 100 characters, it's a potential match for an unknown file that's between 95 and 105 characters. This is important if you have an unknown file that's 1000 characters, that 100 character license file could be swallowed in the comparision if it contained similar words.


mlinksva:
@benbalter :+1: all around.

(1) increase similarity threshold is not needed to fix #111, right? Is 90% causing other false positives?

benbalter:
> (1) increase similarity threshold is not needed to fix #111, right? Is 90% causing other false positives?

It's actually needed for #111. This is a side effect of the Dice matcher, which is fast, but not very smart.

If you take the MIT license and the GPL license and combine them separated by a new line, the resulting license is 98% similar to the GPL license. The MIT license is only 5% the length of the GPL, and about there's about 3% overlap, in terms of the words used in the MIT and the GPL. Put another way, all the words in the MIT but not in the GPL only represent 2% of all words used in the GPL, meaning the Dice matcher would find > 98% confidence, if we didn't first look at file length (which the confidence threshold also determines).

mlinksva:
Gotcha. :+1: :+1:

PR #113, opened 2016-09-14
Special case LICENSE.lesser to handle LGPL implementation instructions

benbalter:
This PR Fixes https://github.com/benbalter/licensee/issues/108.

For better or worse the LGPL FAQ instructs users to add the standard GPL as `LICENSE` and the LGPL as `LICENSE.lesser`. This PR special cases the LGPL. The logic now is as follows:

1. If the LICENSE file isn't a GPL variant, don't do anything special
2. If the LICENSE is a GPL variant, check for a LICENSE.lesser file
3. If the LICENSE.lesser file has a detected license, use that instead, otherwise use the first license

To do this, I abstracted the `find_file` method to be `find_file`, `find_files`, and `files` methods that could be shared between filesystem and Git projects.

/cc @mlinksva for :eyes: on the behavior (which primarily lives now in `project.rb`).


lib/licensee/project_files/license_file.rb
```diff
@@ -20,6 +20,11 @@ def self.name_score(filename)
         return 0.5 if filename =~ /licen[sc]e/i
         0.0
       end
+
+      # case-insensitive block to determine if the given file is LICENSE.lesser
+      def self.lesser_gpl_score(filename)
+        filename.casecmp('license.lesser').zero? ? 1 : 0
```
mlinksva: Will this miss COPYING.LESSER? I suspect that's common, given tradition adhered to by many _GPL'd projects of using COPYING rather than LICENSE.

benbalter: > If you are releasing your program under the LGPL, you should also include the text version of the LGPL, usually in a file called COPYING.LESSER.

Eek. You're correct. Will update the logic and test.

benbalter: Fixed via 3a40440.

mlinksva: FWIW https://github.com/search?utf8=%E2%9C%93&q=filename%3Alicense.lesser finds 1519 results. 10x less than copying.lesser, but _maybe_ enough to handle. licence.lesser only 44.

mlinksva: & copying.lesser.txt 2248 results.

benbalter: Valid point. Although I'm not a fan of their implementation, I'm fine supporting LGPL's recommended means of manifesting your intent to license under LGPL. I'm less keen to support non-standard implementations, for the same reason we don't support `MIT.txt` or `LEGAL-RIGHTS.dog`. We can use proper detection as a carrot to motivate standardization.

mlinksva: Fair enough. I'll remove the "or LICENSE.lesser" from http://choosealicense.com/licenses/lgpl-3.0/ which seems to have been there from the beginning of the site https://github.com/github/choosealicense.com/commit/39198c0b320d97abffcdbf912bcd348a15de6577#diff-c84257b1f58e7d12bce47aed7f0746b0R208

PR #125, opened 2016-09-22
Break license regex into constants

benbalter:
This PR is a refactor intended to DRY up the license file regex a bit.

Rubocop is complaining that `LicenseFIle#name_score` is too complex, and I tend to agree.

Rather than using a series of if statements, this PR moves the logic to iterate through a hash of regexes.

As I did that, I also realized there was a lot of copy-🍝  going on, and DRY'd things up a bit by moving the repeated regex clauses to their own constants.

The end result should be functionally the same, but in my opinion, a cleaner implementation, and should be easier to maintain going forward.

### DIFF
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -5,10 +5,10 @@ module Licensee
 
       # List of extensions to give preference to
       PREFERRED_EXT = %w(md markdown txt).freeze
-      PREFERRED_EXT_REGEX = "\.#{Regexp.union(PREFERRED_EXT)}".freeze
+      PREFERRED_EXT_REGEX = /\.#{Regexp.union(PREFERRED_EXT)}\z/
 
       # Regex to match any extension
-      ANY_EXT_REGEX = /\.[^.]+/
+      ANY_EXT_REGEX = %r{\.[^./]+\z}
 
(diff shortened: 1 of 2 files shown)


lib/licensee/project_files/license_file.rb
```diff
@@ -3,6 +3,32 @@ class Project
     class LicenseFile < Licensee::Project::File
       include Licensee::ContentHelper
 
+      # List of extensions to give preference to
+      EXTENSIONS = %w(md markdown txt).freeze
+      EXTENSIONS_REGEX = Regexp.union(EXTENSIONS)
+
+      # Regex to match any extension
+      ANY_EXTENSION_REGEX = /[^.]+/
```
waldyrious: I'd rename `EXTENSONS_REGEX` to `FORMAT_EXTENSION_REGEX` or `TEXT_EXTENSION_REGEX`, to make the composition below clearer and more consistent.

waldyrious: Ok, I see that it's not just about the format of the file, but actual preferences for it. In that case, I would perhaps use `PREFERRED_EXTENSION_REGEX`, to match `ANY_EXTENSION_REGEX`.

benbalter: Great suggestion.

PR #129, opened 2016-09-23
Add support for the CRAN package manager for R packages.

jimhester:
This PR adds support for detecting license information for R packages (the [CRAN](cran.r-project.org) package repository).

Cran policy, and R package policy in general (see [Licensing](https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Licensing) in the R extensions manual) is that the license contents for standard licenses should _not_ be included in the package. 

In addition for templated licenses (such as MIT) CRAN requires authors to use a LICENSE file with _only_ the follwing (see https://www.r-project.org/Licenses/MIT for the exact instructions)
```
YEAR: 1234
COPYRIGHT HOLDER: XYZ
```

These requirements mean that most R packages are found to be unlicensed by licensee (and consequently GitHub).

This PR adds support for parsing the DESCRIPTION file included in every R package to find the `License:` field. This gets us part of the way to solving the problem.

However because R mandates a restricted LICENSE file for templated licenses and licensee does not use the project matchers if it finds a LICENSE file this PR as written is not sufficient to fix the problem. Licensee would need to fall back to trying the project matchers if no license is found with the normal matching. However as this is a change in the current behavior I did not want to add it to this PR without discussion.

### DIFF
--- a/docs/customizing.md
+++ b/docs/customizing.md
@@ -14,7 +14,7 @@ LICENSEE.confidence_threshold = 90
 
 ## Matching package manager metadata
 
-Licensee supports the ability to take into account Ruby or Node package manager metadata, disabled by default. [There are reasons you may not want to use this metadata](what-we-look-at.md). You can explicitly instruct licensee to take this information into account as follows:
+Licensee supports the ability to take into account Ruby, Node and CRAN package manager metadata, disabled by default. [There are reasons you may not want to use this metadata](what-we-look-at.md). You can explicitly instruct licensee to take this information into account as follows:
 
 ```ruby
 project = Licensee.project("path/to/project", detect_packages: true)
--- a/lib/licensee.rb
+++ b/lib/licensee.rb
@@ -20,6 +20,7 @@ require_relative 'licensee/matchers/dice_matcher'
 require_relative 'licensee/matchers/package_matcher'
 require_relative 'licensee/matchers/gemspec_matcher'
 require_relative 'licensee/matchers/npm_bower_matcher'
+require_relative 'licensee/matchers/cran_matcher'
 
 module Licensee
   # Over which percent is a match considered a match by default
--- a/lib/licensee/matchers/cran_matcher.rb
+++ b/lib/licensee/matchers/cran_matcher.rb
@@ -1,0 +1,18 @@
+module Licensee
+  module Matchers
+    class Cran < Package
+      # While we could parse the package.json or bower.json file, prefer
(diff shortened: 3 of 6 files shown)


benbalter:
@jimhester This is great, and I'd love to get it merged in.

> However because R mandates a restricted LICENSE file for templated licenses and licensee does not use the project matchers if it finds a LICENSE file this PR as written is not sufficient to fix the problem. Licensee would need to fall back to trying the project matchers if no license is found with the normal matching. However as this is a change in the current behavior I did not want to add it to this PR without discussion.

Hmm. That's not the intended behavior, or at least, it's not the ideal. If we can't detect a license via the LicenseMatcher, we should fall back to the PackageMatcher, even if a LICENSE file exists. 

Likewise, we'll also want to confirm that we fallback to look at, e.g., `package.json`, even if an unmatched DESCRIPTION file exists.

Last, is there anything else to do to confirm the project is actually R, and it's not a random description file?

I think it makes the most sense to get this merged, and then can update the logic in a subsequent pass to fallback to the PackageMatcher even when a LICENSE is found but unmatched.

jimhester:
> Last, is there anything else to do to confirm the project is actually R, and it's not a random description file?

I don't think there is any completely deterministic way to do so. R packages can (theoretically) contain no R code, so you cannot check for an `R/` directory. The `DESCRIPTION` files do need to start with `Package: XYZ`, but other fields are optional to varying degrees. Checking the file starts with `Package:` and contains `License:` is likely sufficient in practice, I have not seen any false positives with that criteria by sampling ([filename:"DESCRIPTION" path:/ "Package:"](https://github.com/search?p=94&q=filename%3A%22DESCRIPTION%22+path%3A%2F+%22Package%3A%22&ref=searchresults&type=Code&utf8=%E2%9C%93)). I added this check in f7cb765.

benbalter:
LGTM. Will let this sit for a bit in case anyone else wants to provide feedback.

benbalter:
@jimhester thanks for this. Added a few additional tests and merged via command line.

jeroenooms:
Has this been deployed yet?

benbalter:
Not yet, time permitting, I can try to cut a release this weekend.

jeroenooms:
That would be great!

jimhester:
FWIW this PR isn't that useful for most R packages without the additional changes from https://github.com/benbalter/licensee/pull/160 as well...

PR #130, opened 2016-09-23
Fix a couple of errors in the customizing documentation

jimhester:
1. Broken links.
2. Examples did not work as written.


benbalter:
Nice catch. Technically `project = Licensee::FSProject.new("path/to/project", detect_packages: true)` or `project = Licensee::GitProject.new("path/to/project", detect_packages: true)` should work, but this feels simpler.

PR #131, opened 2016-09-23
Ignore complexity warnings for name_score

jimhester:
I don't think this can complexity can be avoided.

Should fix the CI rubocop errors.


jimhester:
Ah I had missed #125, closing!

benbalter:
Thanks @jimhester!

PR #138, opened 2016-09-26
Recognize underlined license headers in READMEs

pchaigno:
This pull request adds support for underlined headers in READMEs:
```markdown
License
-------

License body
```
It seems that underlined headers are less common than hashes-based headers, but at the scale of github.com, I guess there are still enough to mandate support. It actually improves detection in a few cases for Linguist.

The regular expression gets fairly more complicated so I broke it down into several commented parts with free-spacing.

### DIFF
--- a/lib/licensee/project_files/readme.rb
+++ b/lib/licensee/project_files/readme.rb
@@ -6,7 +6,17 @@ module Licensee
         /\AREADME\.(md|markdown|mdown|txt)\z/i => 0.9
       }.freeze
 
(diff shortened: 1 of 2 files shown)


benbalter:
Tests look solid. /cc @bkeepers on the regex, who originally wrote the matcher. May have some suggestions to make the implementation a bit more readable (e.g., break out the regex into two or more constants that are mixed together), but generally 👍. Nice work.

pchaigno:
@benbalter Is there anything I need to do before this is merged?

benbalter:
Thanks @pchaigno!

PR #140, opened 2016-10-03
reorder docs in a more logical sequence

waldyrious:
use --> customize --> contribute seems to be a much more likely path for users of this tool, than the opposite.


benbalter:
Thanks @waldyrious!

PR #142, opened 2016-10-10
link to RubyGems for those who don't have it :)

waldyrious:


waldyrious:
Actually, this [may not be enough](http://stackoverflow.com/questions/18908708/installing-ruby-gem-in-windows). Any thoughts on what link would be more appropriate to ensure all dependencies are clear to users?

benbalter:
Theoeretically, `gem licensee` should install all necessary dependencies, is thatt not the case?

waldyrious:
I meant the dependencies for the installation itself, i.e. ruby and RubyGems. Of course, the project description states that it is "A Ruby Gem...", but in the installation section it could be worth at least mentioning that installing this requires ruby and RubyGems.

Maybe linking to both https://www.ruby-lang.org/en/downloads/ and https://rubygems.org/pages/download in the installation section could address this. Maybe something like this?

> ## Installation
>
> _(**Note:** To install licensee, you need [Ruby](https://www.ruby-lang.org/en/downloads/) and [RubyGems](https://rubygems.org/pages/download).)_
>
> `gem install licensee` or add `gem 'licensee'` to your project's `Gemfile`.

PR #150, opened 2016-12-24
Add support to match copyright with a full sentence

pablrod:
Some projects have a full sentence for the copyright. For example this [LICENSE file](https://github.com/moose/Class-Load/blob/master/LICENSE) (It's common in Perl modules, specially those minted with [Dist::Zilla](https://metacpan.org/pod/Dist::Zilla))

When this sentence is not matched by the copyright matcher, the license in LICENSE file it's harder to guess for licensee. If this sentence is treated as the copyright licensee can do a better work guessing the license.

### DIFF
--- a/lib/licensee/matchers/copyright_matcher.rb
+++ b/lib/licensee/matchers/copyright_matcher.rb
@@ -5,7 +5,7 @@ module Licensee
       attr_reader :file
 
       # rubocop:disable Metrics/LineLength
-      REGEX = /\s*(Copyright|\(c\)) (©|\(c\)|\xC2\xA9)? ?(\d{4}|\[year\])(.*)?\s*/i
+      REGEX = /\s*(This software is )?([Cc]opyright|\(c\)) (©|\(c\)|\xC2\xA9)? ?(\d{4}|\[year\])(.*)?\s*/i
 
       def initialize(file)
         @file = file
--- a/spec/licensee/matchers/copyright_matcher_spec.rb
+++ b/spec/licensee/matchers/copyright_matcher_spec.rb
@@ -27,7 +27,8 @@ RSpec.describe Licensee::Matchers::Copyright do
     'Hyphen-separated date' => 'Copyright (c) 2003-2004 Ben Balter',
     'ASCII-8BIT encoded'    => "Copyright \xC2\xA92015 Ben Balter`"
       .force_encoding('ASCII-8BIT'),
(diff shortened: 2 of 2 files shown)


lib/licensee/matchers/copyright_matcher.rb
```diff
@@ -5,7 +5,7 @@ class Copyright
       attr_reader :file
 
       # rubocop:disable Metrics/LineLength
-      REGEX = /\s*(Copyright|\(c\)) (©|\(c\)|\xC2\xA9)? ?(\d{4}|\[year\])(.*)?\s*/i
+      REGEX = /\s*(This software is )?([Cc]opyright|\(c\)) (©|\(c\)|\xC2\xA9)? ?(\d{4}|\[year\])(.*)?\s*/i
```
benbalter: Since the regex has the `/i` flag, I don't believe the `[Cc]` is necessary.

benbalter:
I don't have a strong objection, to the feature, *per se*, but what concerns me, is [fewer than 40 of GitHub's 40M projects use this syntax](https://github.com/search?p=1&q=%22This+software+is+copyright%22+path%3ALICENSE&ref=searchresults&type=Code&utf8=%E2%9C%93), the majority of which appears to be forks of a single project. 

If that is, in fact, the case, I think the better solution would be to update those few dozen licenses, than to add the complexity here.

lib/licensee/matchers/copyright_matcher.rb
pablrod: Good point! Sorry I missed the ` \i` flag.
Thank you!

pablrod:
Good search! Thank you! 
BTW do you know why [this repo](https://github.com/pablrod/p5-HTML-Notebook/blob/master/LICENSE#L1) is not in that search?

With your search I've realized that most of the projects I was thinking about doesn't have the LICENSE file in the version control. This is because that projects are Perl modules using [Dist::Zilla](http://dzil.org). Dist::Zilla uses a dist.ini file with a license field (more or less like package.json, bower.json or Gemspec). Do you think is better to go that way?

benbalter:
> Do you think is better to go that way?

I think that's the better way to go. There's prior art in https://github.com/benbalter/licensee/tree/master/lib/licensee/matchers, if you're interested in adding a `dist.ini` matcher.

pablrod:
Ok. I'm going to add a `dist.ini` matcher. When it's done I'll open a new pull request.

Thank you!

PR #151, opened 2016-12-27
Add Ruby 2.3 and 2.4 to the test matrix.

connorshea:
Happy holidays!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ rvm:
   - 2.0
   - 2.1
   - 2.2
+  - 2.3.3
+  - 2.4.0
 script: "script/cibuild"
 sudo: false
 cache: bundler


connorshea:
Until rails/rails@47c4924 is released in Rails 4.2.8, Ruby 2.4 won't work with it. The test suite uses activesupport unfortunately.

benbalter:
@connorshea thanks. Could you explain a bit more what you're trying to solve for here? I'm afraid I'm lost.

connorshea:
@benbalter rails 4.2.7 depends on json 1.x, which doesn't work on Ruby 2.4. In order to test on Ruby 2.4 we have to wait for rails 4.2.8 to come out which doesn't depend on json 1.x.

Sorry if that still isn't clear 🙈

benbalter:
@connorshea sorry this never got merged. We're now testing 2.3 and 2.4, but will go ahead and merge this PR so you can get the contribution bit. Thanks for the PR! :smile:

connorshea:
@benbalter no problem, thanks for maintaining the gem :)

PR #152, opened 2016-12-27
Drop active support

benbalter:
In support of https://github.com/benbalter/licensee/pull/151.

It doesn't look like ActiveSupport is used. It was added via https://github.com/benbalter/licensee/pull/107, but at least locally, tests pass without it being bundled. I suspect it was a remnant of an alternate approach to testing.

/cc @connorshea

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,3 @@
 source 'https://rubygems.org'
 
 gemspec
-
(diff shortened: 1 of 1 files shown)


connorshea:
Sounds good to me! :+1:

PR #153, opened 2016-12-28
Match licenses, not their instructions on how to apply them

talisein:
The GPL licenses have 'How to apply to your program' instructions at the bottom of their file. This text is not the license itself, and some people remove the instructions when putting the LICENSE file in their repo.

I've tested this against the GPL--with this pull request, license files that include the instructions are still matched, while license files that do not include instructions are now matched correctly as well.


talisein:
Ah, I see I'll need to fix the vendor-licenses script instead.

mlinksva:
@talisein I probably wouldn't merge this over at github/choosealicense.com which the script pulls from. Deployment without optional instructions seems really uncommon and I wouldn't want to encourage the practice.

Ideally both projects could get and exploit annotated (including with optional sections) license texts, as being developed at https://github.com/spdx/license-list-XML or similar. But that would require huge implementation changes and I imagine is in the far future if ever.

talisein:
I agree choosealicense.com should include the instructions. 
The vendor-licenses script could strip text after 'END OF TERMS AND CONDITIONS' as a post-processing step.

mlinksva:
Ah, my reading of your "fix the vendor-licenses script instead" was insufficiently imaginative _and_ insufficiently literal. :smile:

PR #154, opened 2016-12-28
Only consider license content until 'END OF TERMS AND CONDITIONS'

talisein:
The GPL licenses have 'How to apply to your program' instructions at the bottom of their file. This text is not the license itself, and some people remove the instructions when putting the LICENSE file in their repo.

### DIFF
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -49,6 +49,8 @@ module Licensee
         content_normalized = content.downcase.strip
         content_normalized.gsub!(/^#{Matchers::Copyright::REGEX}$/i, '')
         content_normalized.gsub!(/[=-]{4,}/, '') # Strip HRs from MPL
+        h, m, _t = content_normalized.rpartition(/end of terms and conditions/)
+        content_normalized = h.concat(m) unless h.empty?
         content_normalized.tr("\n", ' ').squeeze(' ').strip
       end
     end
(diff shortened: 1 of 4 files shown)


benbalter:
Thanks @talisein!

@mlinksva any strong feelings one way or the other?

mlinksva:
Seems fine to me so long as it doesn't lead us to stop matching with instructions included. Are there tests for matching both normal and instructions-removed versions of the texts this truncates for matching purposes?

mlinksva:
(Even if they match now, tests are important in case we separately decide to further tweak matching thresholds, etc.)

talisein:
I added a test for matching when the repository under test is GPLv3 with no instructions. There are existing tests where the repository under test includes the instructions (e.g. fixtures/lgpl). If you would like me to create explicit tests for all the licenses this effects today (Apache, AGPL, LGPL, GPL) I don't mind, but it might clutter up the fixtures directory a little.

benbalter:
The technical implementation is fine here (thanks @talisein). Between this and the [Apache boilerplate](https://github.com/benbalter/licensee/issues/103), I'm wondering if licenses (at least those two), should have an alternate form, which we could test. I'd prefer we test the versions that the vast majority use (which could result in an exact match, with 100% confidence) than a mangled version that's still above our 95% threshold (which we may one day want to raise or lower, which could result in false positive/negatives).

benbalter:
>  I'd prefer we test the versions that the vast majority use (which could result in an exact match, with 100% confidence) than a mangled version that's still above our 95% threshold 

@talisein Looking at your implementation again, I changed my mind and like your approach. If a license says "END OF TERMS AND CONDITIONS", it should be safe to ignore anything after it from an enforceability perspective, and as long as we're doing it on both sides, as you are, the result should be the same.

I'm going to resolve the merge conflict and get this merged.

benbalter:
@talisein Thanks for this. Resolved the merge conflict and merged via https://github.com/benbalter/licensee/pull/154.

PR #160, opened 2017-01-18
Complete support for R/CRAN License conventions

jimhester:
This continues work begun with #129. In particular it allows fallback to
the readme and project matchers if the license_file does not contain a
valid license.

We also special case the most common naming conventions with CRAN
packages, as the license matching expects exact name matching, so
without this most of the ly used license terms in R packages would not be
found. The top 6 most common strings on CRAN (with counts) are

3484 GPL (>= 2)
1973 GPL-2
1159 GPL-3
534 MIT + file LICENSE
522 GPL
459 GPL (>= 3)

With these changes many more R package licenses should be properly
detected by Licensee.


jimhester:
The build failures all seem to be rubocop lints. Many are coming from files untouched by the PR, I am not sure what has changed there. There are a couple of cyclomatic complexity warnings from this PR though. If you have any thoughts on a clean way to reduce the cyclomatic complexity while preserving the behavior I am happy to adjust the code accordingly.

spec/fixtures/description-license/DESCRIPTION
```diff
@@ -0,0 +1,2 @@
+Package: xyz
+License: MIT + file LICENSE
```
benbalter: If the author is saying this is licensed under the MIT license + whatever arbitrary terms I put in the license file, why is it accurate to say the project is licensed under the MIT license? (e.g., if the license file says "and you must give me 100 puppies if you use this software")

jimhester: Per CRAN policy MIT is what they call a 'license template' (https://www.r-project.org/Licenses/MIT), which states a structured format for `LICENSE` you need to follow to comply and be accepted into CRAN. Deviations from this form will be rejected. So while in principle what you say is accurate, in practice these packages are using the basic MIT license.

lib/licensee/project_files/package_info.rb
```diff
@@ -8,7 +8,7 @@ def possible_matchers
         when '.json'
           [Matchers::NpmBower]
         else
-          if filename == 'DESCRIPTION' && content.start_with?('Package:')
+          if filename == 'DESCRIPTION' && content.include?('Package:')
```
benbalter: Does the line have to begin `Package?` (or can it be in the middle of a line, even if not the first)?

lib/licensee/matchers/cran_matcher.rb
```diff
@@ -3,17 +3,31 @@ module Matchers
...
       private
 
       def license_property
         match = @file.content.match LICENSE_REGEX
-        match[1].downcase if match && match[1]
+        return unless match && match[1]
+
+        # Remove The common + file LICENSE text
+        match = match[1]
+        match.slice!(/\s*\+\s*file\s*LICENSE.*/)
```
benbalter: Can this be `match[1].gsub(/+ file LICENSE$/,'')`?

lib/licensee/matchers/cran_matcher.rb
```diff
@@ -3,17 +3,31 @@ module Matchers
...
+
+        # Match GPL (>=2)
+        m = match.match(/^GPL\s*\(>=\s*([23])\)/)
+        return "gpl-#{m[1]}.0" if m && m[1]
```
benbalter: Can these two be combined as `/^GPL(-|>=)(\d)/`?

lib/licensee/project_files/package_info.rb
jimhester: The line has to start with package, but it doesn't have to be the first line.

lib/licensee/matchers/cran_matcher.rb
jimhester: Sure

lib/licensee/matchers/cran_matcher.rb
jimhester: Sure

jimhester: Actually no, the two options are [`GPL-2`](https://github.com/cran/ttutils/blob/bd83bbd6b0b7408502a1060da0274255343b9819/DESCRIPTION#L8), or [`GPL (>=2)`](https://github.com/cran/richards/blob/32f323299ab88c879f96d0906747adf98cdbca34/DESCRIPTION#L11)

jimhester:
Ok I tried to address your review comments, thank you for the review and patience!

benbalter:
Thanks @jimhester! Pushed up two commits to make Rubocop (style check) happy and this looks great!

PR #161, opened 2017-01-18
Update range of linked lines in docs

roryokane:
`license_file.rb` was edited since the link was first made, so the range pointed to the wrong part of the file. With this change, the range once again cover the `FILENAME_REGEXES` hash contents.


benbalter:
Thanks @roryokane!

PR #164, opened 2017-02-16
avoid detecting CC -NC or -ND licenses

### LINKED ISSUE #116: CC-BY-ND-4.0 is detected as CC-BY-4.0

mlinksva:
fixes #116

Probably could be thrown out earlier, but seemed most obvious to do where only other special case is handled (for LGPL).


benbalter:
@mlinksva cherry-picked b08c829, 2e04db9, and f5bdca1 onto master to fix tests. Good catch.

PR #165, opened 2017-02-20
docs/what-we-look-at: Remove "Is it GPLv2? GPLv2 or later?"

wking:
Some packaging systems allow you to make these distinctions.  For example, Python supports [“or later” Trove classifiers][1], including `License :: OSI Approved :: GNU General Public License v2 (GPLv2)` and `License :: OSI Approved :: GNU General Public License v2 or later (GPLv2+)`.  npm [allows SPDX license expression syntax v2.0][2] (and [SPDX expressions support `+`, e.g. `GPL-2.0+`][3]).

The point of this paragraph is that such an entry is metadata which may not be sufficiently binding (#4), while including the full license text is much more likely to be legally binding.  By removing the FUD about “cannot distinguish between GPL-2.0 and GPL-2.0+”, we can focus on the point that actually matters.

[1]: https://pypi.python.org/pypi?%3Aaction=list_classifiers
[2]: https://docs.npmjs.com/files/package.json#license
[3]: https://spdx.org/sites/cpstandard/files/pages/files/spdx-2.0.pdf#page=82
[4]: https://github.com/benbalter/licensee/issues/4

### DIFF
--- a/docs/what-we-look-at.md
+++ b/docs/what-we-look-at.md
@@ -33,9 +33,7 @@ Because reasons.
 
 Because it's not legally binding. A license is a legal contract. You give up certain rights (e.g., the right to sue the author) in exchange for the right to use the software.
 
-Most popular licenses today *require* that the license itself be distributed along side the software. Simply putting the letters "MIT" or "GPL" in a configuration file doesn't really meet that requirement.
-
-Not to mention, it doesn't tell you much about your rights as a user. Is it GPLv2? GPLv2 or later? Those files are designed to be read by computers (who can't enter into contracts), not humans (who can). It's great metadata, but that's about it.
+Most popular licenses today *require* that the license itself be distributed along side the software. Simply putting the letters "MIT" or "GPL" in a configuration file doesn't really meet that requirement. Those files are designed to be read by computers (who can't enter into contracts), not humans (who can). It's great metadata, but that's about it.
 
 ### What about looking to see if the author said something in the readme?
 


wking:
The [Travis errors][1] appear unrelated to this small Markdown change ;).

[1]: https://travis-ci.org/benbalter/licensee/jobs/203595172#L230

benbalter:
@wking thanks. I think it's *inaccurate* to say "package managers don't support 'or later' syntax", as your links prove, but in merging this, I don't want to confuse "some package managers support 'or later' syntax" with "just putting the letters 'GPL' in a file is sufficient".

wking:
On Tue, Feb 21, 2017 at 08:50:35AM -0800, Ben Balter wrote:
> … I don't want to confuse "some package managers support 'or later'
> syntax" with "just putting the letters 'GPL' in a file is
> sufficient".

There are a few levels to this (as you've sketched out in #4).  For
the GPL in particular, versions 1, 2, and 3 all have wording around
“If the Program does not specify a version number” [2], so declaring
‘GPL’ is as clear as declaring ‘GPL-2.0’, ‘GPL-3.0+’, etc.  This may
not be true of other licenses.  So I see the following clarity
groupings:

a. Just a metadata string
  * GPL (all current GPL versions explain this to be GPL-1.0+, but
    maybe the author was being careless)
  * GPL-2.0 (literally means “just this license”, but maybe the author
    was being careless and actually wanted GPL-2.0+)
  * GPL-2.0+ (very clearly means “v2.0 or later”.  Unlikely to be an
    accidental mistake)
b. Just the full text of a license
  * Can only mean “just this license”, but maybe the author was being
    careless and actually wanted GPL-2.0+.
c. Full text of a license and a metadata string or native-language
   description
  * E.g. full GPL-2.0 text plus GPL-2.0+ in package.json
  * E.g. full GPL-2.0 text plus “… you can redistribute it and/or modify
    it under the terms of the GNU General Public License as published by
    the Free Software Foundation, either version 2 of the License, or
    (at your option) any later version.”

In this commit, I dropped the previous attempt at distinguishing
within (a) to focus on the distinction between (a) and (b)/(c).  So
“you have to include the full license text, even if your metadata
string is a really clear SPDX License Expression”.

I agree with you that “just putting the letters ‘GPL’ in a file is not
sufficient”.  But I also think that “just putting the letters
'GPL-3.0+' in a file is not sufficient”, and makes the point we're
trying to make more obvious.

[1]: https://github.com/benbalter/licensee/issues/4#issuecomment-155195954

PR #166, opened 2017-02-21
Avoid detecting CC-NC and CC-ND as false positives for CC-BY and CC-BY-SA

benbalter:
@mlinksva this PR builds on your work over in https://github.com/benbalter/licensee/pull/164.

I took your initial commit and then took a slightly different approach. Specifically, I created a `FalsePositive` matcher, for two reasons:

1. I think we should push the detection logic as far down into the matchers as possible, rather than matchers returning bad results and handling them on e.g., the project level (which would create a confusing API experience for those in Ruby land).

2. Return the `other` license, rather than `nil` for licenses that we know have a license file, but are affirmatively detected as a non-open-source license. This keeps consistency of experience with `nil` representing "no license file" and `other` representing "license we know is there but don't detect". 

This approach also allows for tighter unit tests.

To implement the `FalsePositive` matcher, it looks for a license that begins with either NC or ND's title. As that a safe assumption? A quick GitHub code search suggested that was the case (and I'm trying to avoid the FalsePositive matcher itself creating False negatives).

Let me know if this approach makes sense here (and can then be theoretically applied to other known false positives based on regex).

### DIFF
--- a/lib/licensee.rb
+++ b/lib/licensee.rb
@@ -23,6 +23,7 @@ require_relative 'licensee/matchers/gemspec_matcher'
 require_relative 'licensee/matchers/npm_bower_matcher'
 require_relative 'licensee/matchers/cran_matcher'
 require_relative 'licensee/matchers/dist_zilla_matcher'
+require_relative 'licensee/matchers/false_positive_matcher'
 
 module Licensee
   # Over which percent is a match considered a match by default
--- a/lib/licensee/matchers/false_positive_matcher.rb
+++ b/lib/licensee/matchers/false_positive_matcher.rb
@@ -1,0 +1,29 @@
+module Licensee
+  module Matchers
+    class FalsePositive
+      attr_reader :file
+
+      # CC-NC and CC-ND are not open source licenses and should always be
+      # detected as the "other" license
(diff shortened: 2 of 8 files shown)


mlinksva:
@benbalter LGTM. A few brief notes:

- Returning 'other' is definitely the right thing to do, I overlooked that
- Agree on pushing detection as far down as possible. I didn't start that way but did in https://github.com/benbalter/licensee/pull/164/commits/72c338aa81161a4c2b44564a525e07b51ea1db19
- Only downside here IIUC is that the `FalsePositive` matcher will get run always when there isn't an exact match; commit above tries to only test for false positive when we have a license that we have a false positive test for.
- Title assumption is probably safe. The only thing I wonder (I did not check) is whether a license with some or all of the title deleted would still match CC-BY[-SA]. That's why I just checked for `NonCommercial` or `NoDeriv`.
- Agree completely that this mechanism could be useful for detecting other false positive situations if we discover them in the future.

benbalter:
@mlinksva your rationale makes sense. I might make a small style change before merging, but I'm going to close this PR in favor of your implementation over in #164. Thanks for talking through things.

PR #168, opened 2017-02-21
Third attempt at avoid detecting CC-NC and CC-ND as false positives for CC-BY and CC-BY-SA 

benbalter:
A follow up to https://github.com/benbalter/licensee/pull/166 and https://github.com/benbalter/licensee/pull/164, this PR builds on @mlinksva's work, but simply moves the CC false detection logic into the dice matcher (where the false positive lives), rather than in the content helper (which can't assume the thing being compared has a `key` as it may be a license file).

This way, we avoid the strange `license.license_similarity(other)` in favor of `license.similiarity(other)`, but the list of `potential_licenses` now excluded CC licenses if the license file looks like a potential CC false positive.

To do this, I added a `creative_common?` helper to the license model (to match the existing GPL helper), and `potential_false_positive?` to LicenseFile to implement @mlinksva title detection logic.

Last, since this is now broken out into distinct methods, I added unit tests for each (in addition to @mlinksva's integration tests).

### DIFF
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -29,8 +29,8 @@ module Licensee
         //                                           => 0.0  # Catch all
       }.freeze
 
-      # CC-NC and CC-ND are not open source licenses and should always be
-      # detected as the "other" license
+      # CC-NC and CC-ND are not open source licenses and should not be
+      # detected as CC-BY or CC-BY-SA which are 98%+ similar
       CC_FALSE_POSITIVE_REGEX = /
         \A(creative\ commons\ )?Attribution-(NonCommercial|NoDerivatives)
       /xi
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -29,6 +29,12 @@ module Licensee
         //                                           => 0.0  # Catch all
       }.freeze
 
+      # CC-NC and CC-ND are not open source licenses and should always be
+      # detected as the "other" license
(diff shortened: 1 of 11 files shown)


lib/licensee/project_files/license_file.rb
```diff
@@ -29,6 +29,12 @@ class LicenseFile < Licensee::Project::File
         //                                           => 0.0  # Catch all
       }.freeze
 
+      # CC-NC and CC-ND are not open source licenses and should always be
+      # detected as the "other" license
```
mlinksva: Are they? The tests are still expecting nil.

mlinksva:
> list of `potential_licenses` now excluded CC licenses if the license file looks like a potential CC false positive.

Yet another approach! I like it.

lib/licensee/project_files/license_file.rb
benbalter: Clarified via 3579d6f. I was wrong. In Licensee, the existence of a LicenseFile shows that there is a license. If no matcher returns a match (e.g., DiceMatcher returns nil for match), that means the license hasn't been detected (but that there is a license file). In the GitHub API, we go a step further and collapse that into `other`, nil, or the detected license.

benbalter:
Just noting for posterity, a fourth approach could be to require exact matches (once normalized) since CC has no fillable fields, but I feel like that's too fragile for real-world use. Going to go with this approach for now so we can cut a release (and since we should have some flexibility within the API to make changes to the logic, if need be).

PR #176, opened 2017-03-20
Look up license rule by tag and group

benbalter:
This fixes the case of `patent-use` under limitations for CC licenses showing the wrong description by looking up license rules by both tag and group, not just tag. 

We previously assumed rule tags were globally unique. They're only unique to the group.

We also weren't running the rule spec apparently due to the file name (although tests mostly passed).

### DIFF
--- a/lib/licensee/rule.rb
+++ b/lib/licensee/rule.rb
@@ -27,9 +27,10 @@ module Licensee
         end.flatten
       end
 
-      def find_by_tag(tag)
-        Rule.all.find { |r| r.tag == tag }
+      def find_by_tag_and_group(tag, group = nil)
+        Rule.all.find { |r| r.tag == tag && (group.nil? || r.group == group) }
       end
+      alias find_by_tag find_by_tag_and_group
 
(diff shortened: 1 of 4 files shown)


lib/licensee/rule.rb
```diff
@@ -27,9 +27,10 @@ def all
         end.flatten
       end
 
-      def find_by_tag(tag)
-        Rule.all.find { |r| r.tag == tag }
+      def find_by_tag_and_group(tag, group = nil)
+        Rule.all.find { |r| r.tag == tag && (group.nil? || r.group == group) }
       end
+      alias find_by_tag find_by_tag_and_group
```
mlinksva: Just curious, are the alias and default nil value for group above for backwards compatibility?

benbalter: Yep. Trying to follow http://semver.org/ strictly since https://github.com/benbalter/licensee/pull/100.

This way, if you call `Rule#find_by_tag('foo')` it still works without needing to maintain two different methods.

PR #179, opened 2017-03-28
Strip HRs before looking for title and copyright

benbalter:
This solves the case of:

```
The MIT License
============

Copyright 2016...
```


waldyrious:
I'd suggest removing `#`s before the title as well, since it's also used to format headings.

PR #182, opened 2017-03-31
Resolve lint errors

willbar:
This resolves the current build failure. I'm new to Ruby and resolving the requirement for heredoc to be indented wasn't pretty, pulling in a new dependency seemed like overkill so used the ```gsub``` solution from http://stackoverflow.com/questions/3772864/how-do-i-remove-leading-whitespace-chars-from-ruby-heredoc.


benbalter:
Thanks @willbar! 🎉 💥 ✨

PR #186, opened 2017-04-10
Expanded support for standard files declaring the font under OFL. 

n7s:
Many open fonts developed and released under OFL on GitHub are not detected properly by licensee.  A key feature of the license was naming the license file explicitly (instead of using the generic software filenames) so designers could recognize it more easily. 

Best practices across the open font design community is to use OFL.txt. 
Please add support for the standard filenames. 

I can see you have added already support the standards files like MIT.txt or MIT-license.txt. This is similar. 

Thanks.

### DIFF
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -17,22 +17,22 @@ module Licensee
       COPYING_REGEX = /copy(ing|right)/i
 
       # Regex to match OFL.
-      OFL_REGEX = /OFL(.txt|.md)/i
+      OFL_REGEX = /ofl/i
 
       # Hash of Regex => score with which to score potential license files
       FILENAME_REGEXES = {
-        /\A#{LICENSE_REGEX}\z/                       => 1.0, # LICENSE
-        /\A#{LICENSE_REGEX}#{PREFERRED_EXT_REGEX}\z/ => 0.9, # LICENSE.md
-        /\A#{COPYING_REGEX}\z/                       => 0.8, # COPYING
-        /\A#{COPYING_REGEX}#{PREFERRED_EXT_REGEX}\z/ => 0.7, # COPYING.md
-        /\A#{LICENSE_REGEX}#{ANY_EXT_REGEX}\z/       => 0.6, # LICENSE.textile
-        /\A#{COPYING_REGEX}#{ANY_EXT_REGEX}\z/       => 0.5, # COPYING.textile
-        /#{LICENSE_REGEX}/                           => 0.4, # LICENSE-MIT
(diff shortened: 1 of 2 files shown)


n7s:
This also add support for OFL.md

benbalter:
@n7s thanks for the PR. If we did this, I wouldn't want the implementation to be specific to OFL. We don't currently support `MIT.md` as far as I know, but I don't have a strong objection to matching `SPDX-ID-spit-at-the-first-hyphen.whatever` as a last resort. @mlinksva any strong feelings either way or prior art?

mlinksva:
I tend to think only supporting OFL might be the right thing to do. The practice does appear to be pretty widespread -- https://github.com/search?q=filename%3AOFL+%22open+font+license%22 gets 90k results (though many may be fonts in subdirectories). I don't know that bare SPDX-ID-sans-version[.txt|md] is a common practice for any other license, and assuming it is not, I don't think we'd want to encourage it.

n7s:
AFAICT a majority of font designers don't (yet) have a very deep culture of software development or want to care too deeply about the whole spectrum of licenses. Calling all licenses COPYING (or the generic LICENSE) was a recipe for confusion or indifference...  Seeing and using the template OFL.txt seems to work best for them in their own projects. GitHub supporting these standards would be great (along with secondary support for OFL - no extension - and OFL.md). Thanks a lot.

mlinksva:
It's really too bad that they didn't adopt `LICENSE-OFL` as a practice. `OFL` doesn't mean anything at all to anyone who isn't involved in fonts.

lib/licensee/project_files/license_file.rb
```diff
@@ -26,6 +30,7 @@ class LicenseFile < Licensee::Project::File
         /\A#{COPYING_REGEX}#{ANY_EXT_REGEX}\z/       => 0.5, # COPYING.textile
         /#{LICENSE_REGEX}/                           => 0.4, # LICENSE-MIT
         /#{COPYING_REGEX}/                           => 0.3, # COPYING-MIT
+        /#{OFL_REGEX}/                               => 0.5, # OFL
```
benbalter: To match the other standards, I believe we'll want `/OFL(#{PREFERRED_EXT_REGEX}` and `/OFL#{ANY_EXT_REGEX}/` as their own entries. It should also be scored lower than `COPYING-*`. when both exist.

n7s:
Anything I can do to adjust this further? Thanks.

benbalter:
@n7s Thanks for this. I updated the regex to match the other file regexes via cacdda0, and then added tests via a5fbd7c.

PR #189, opened 2017-04-21
Updating vendored licenses from the choosealicense repo 

narendasan:
There were a couple new licenses added to the choosealicense repo, this pr should bring licensee up to date with it


benbalter:
Thanks @narendasan!

PR #198, opened 2017-06-05
Don't allow any match to differ by > 150 characters

### LINKED ISSUE #192: Modified Apache 2.0 License is detected as default Apache
See https://github.com/wg/wrk/blob/master/LICENSE

Github displays this as the default "Apache License 2.0" even though the title of the license clearly says "Modified", and there are additional terms within the license text.  This can result in users being surprised later on if the author attempts to exert their rights from the additional clause, because with how Github makes their "_project_ is licensed under the _X_ license", many users may not see the change in the license text.

Thanks for considering!

mlinksva:
Fixes #192 taking 3rd approach I suggested:

> Different attack on false positives for long licenses without changing % words in common: reject match if absolute number of words is different; again getting this right would require fiddling

150 allows all existing tests to pass (much less than that MPL and LPPL rewrap tests fail).

This fixes a couple known false positives involving modifications of long licenses, which I've included as integration tests.

There are doubtless more robust or more principled approaches that should be explored longer term.

### DIFF
--- a/lib/licensee.rb
+++ b/lib/licensee.rb
@@ -55,7 +55,7 @@ module Licensee
     end
 
     # Inverse of the confidence threshold, represented as a float
-    # By default this will be 0.1
+    # By default this will be 0.05
     def inverse_confidence_threshold
       @inverse ||= (1 - Licensee.confidence_threshold / 100.0).round(2)
     end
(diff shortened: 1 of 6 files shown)


clr-apache:
Just one minor comment. Sometimes projects use the entire Apache 2.0 license including the Appendix: How to apply the license to your own work. And sometimes projects just use the license ending with END OF TERMS AND CONDITIONS. 

If you are already aware, we're done here.

benbalter:
> And sometimes projects just use the license ending with END OF TERMS AND CONDITIONS.

[We strip the Apache appendix in all cases](https://github.com/benbalter/licensee/blob/master/lib/licensee/content_helper.rb#L75) when we normalize the input.

clr-apache:
Excellent. Thanks.

PR #199, opened 2017-06-06
Use Charlock Holmes for encoding detection/conversion

### LINKED ISSUE #79: Mangles attribution line by stripping non-ASCII characters
For example, given a `COPYING` file with contents

```
Copyright © 2013–2016 by Peder Ås, 王二麻子, and Seán Ó Rudaí
```

`licensee` returns

```
License file: COPYING
Attribution: Copyright  20132016 by Peder s, , and Sen  Ruda
License: No License
Confidence: 100%
Method: Licensee::Matchers::Copyright
```

mlinksva:
Fixes #79 

Introduces a new dependency, found at https://github.com/libgit2/rugged/issues/588#issuecomment-202773337 which is probably [a pain install on Windows](https://github.com/brianmario/charlock_holmes/issues/84#issuecomment-140239662).

I have no opinion on whether worth trouble.

### DIFF
--- a/lib/licensee/project_file.rb
+++ b/lib/licensee/project_file.rb
@@ -1,18 +1,18 @@
+require 'charlock_holmes'
+
 module Licensee
   class Project
     class File
       attr_reader :content, :filename
 
-      ENCODING = Encoding::UTF_8
-      ENCODING_OPTIONS = {
-        invalid: :replace,
-        undef:   :replace,
-        replace: ''
-      }.freeze
-
       def initialize(content, filename = nil)
         @content = content
-        @content.encode!(ENCODING, ENCODING_OPTIONS)
+        unless @content.empty?
+          detection = CharlockHolmes::EncodingDetector.detect(content)
+          @content = CharlockHolmes::Converter.convert @content,
+                                                       detection[:encoding],
+                                                       'UTF-8'
+        end
         @filename = filename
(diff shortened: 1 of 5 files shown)


benbalter:
@vmg Do you have any strong feelings about adding Charlock Holmes as a dependency of Licensee?

vmg:
The Rails app already depends on it, so from that point of view, that won't be an issue.

For random external users, however, this is totally gonna suck. :)

lib/licensee/project_file.rb
```diff
@@ -1,18 +1,18 @@
...
-      ENCODING_OPTIONS = {
-        invalid: :replace,
-        undef:   :replace,
-        replace: ''
-      }.freeze
-
       def initialize(content, filename = nil)
         @content = content
-        @content.encode!(ENCODING, ENCODING_OPTIONS)
+        unless @content.empty?
```
vmg: I think we need to do better here. Charlock is not free (particularly not the transcoding step). The logic needs to be more like:

```ruby
@content.force_encoding(Encoding::UTF_8)
if !@content.valid_encoding?
  # do the transcoding
end
```

mlinksva: That was the hint I needed to avoid Charlock entirely. Thanks! Alternative in #200

PR #200, opened 2017-06-07
Try forcing forcing encoding before transcoding

### LINKED ISSUE #79: Mangles attribution line by stripping non-ASCII characters

mlinksva:
Fixes #79 without new dependency of #199

### DIFF
--- a/lib/licensee/project_file.rb
+++ b/lib/licensee/project_file.rb
@@ -12,7 +12,10 @@ module Licensee
 
       def initialize(content, filename = nil)
         @content = content
-        @content.encode!(ENCODING, ENCODING_OPTIONS)
+        @content.force_encoding(ENCODING)
+        unless @content.valid_encoding?
+          @content.encode!(ENCODING, ENCODING_OPTIONS)
+        end
         @filename = filename
       end
 
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -5,7 +5,7 @@ AllCops:
     - test/fixtures/**/*
     - test/functions.rb
 
-Style/AlignHash:
+Layout/AlignHash:
   EnforcedHashRocketStyle: table
   EnforcedColonStyle: table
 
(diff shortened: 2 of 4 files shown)


lib/licensee/project_file.rb
```diff
@@ -12,7 +12,10 @@ class File
 
       def initialize(content, filename = nil)
         @content = content
-        @content.encode!(ENCODING, ENCODING_OPTIONS)
+        @content.force_encoding(ENCODING)
+        unless @content.valid_encoding?
```
vmg: This is pretty good. Do note that encoding an UTF8-tagged script into UTF-8 with `:replace` will remove all the invalid characters altogether from the License -- as opposed to to Charlock's approach that actually figures out the encoding of those characters and transcodes them into valid UTF8...

So we're gonna be losing data here, and hence reducing accuracy. But I think this is an acceptable tradeoff, particularly since the text of all our supported licenses is usually ASCII English and this approach will only strip away things like author names.

mlinksva: > `:replace:` will remove all the invalid characters altogether from the License

Yes, that's exactly what was causing #79.

Trying force_encoding first, and only running encode (which does above) if that doesn't produce a valid encoding should avoid the latter in many cases. (I didn't realize you could check for a valid encoding, that was the hint I needed.)

vmg: Fantastic. Let's merge!

benbalter:
Thanks @vmg, @mlinksva!

PR #201, opened 2017-06-08
Add support for determining the license from an arbitrary URL

willbar:
Allows you to pass an arbitrary URL to determine the license for the content at that URL. I've also resolved all of the current lint warnings.

If you want me to remove the .vscode/launch.json I can, it just enables easy debugging for anyone using Visual Studio Code.

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -63,7 +63,7 @@ module Licensee
     # `no-license` - The project is not licensed (e.g., all rights reserved)
     #
     # Note: A lack of detected license will be a nil license
-    PSEUDO_LICENSES = %w(other no-license).freeze
+    PSEUDO_LICENSES = %w[other no-license].freeze
 
     include Licensee::ContentHelper
 
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -63,7 +63,7 @@ module Licensee
     # `no-license` - The project is not licensed (e.g., all rights reserved)
     #
     # Note: A lack of detected license will be a nil license
-    PSEUDO_LICENSES = %w[other no-license].freeze
+    PSEUDO_LICENSES = %w(other no-license).freeze
 
     include Licensee::ContentHelper
 
--- a/lib/licensee/projects/uri_project.rb
+++ b/lib/licensee/projects/uri_project.rb
@@ -7,7 +7,8 @@ require 'uri'
 module Licensee
   class UriProject < Project
     def initialize(path, **args)
-      @uri = path
+      @uri = URI(path)
+      raise UnsupportedProject if @uri.scheme != 'http' and @uri.scheme != 'https'
       super(**args)
     end
 
--- a/lib/licensee/projects/uri_project.rb
+++ b/lib/licensee/projects/uri_project.rb
@@ -8,7 +8,7 @@ module Licensee
   class UriProject < Project
     def initialize(path, **args)
       @uri = URI(path)
-      raise UnsupportedProject if @uri.scheme != 'http' and @uri.scheme != 'https'
+      raise UnsupportedProject if @uri.scheme != 'http' && @uri.scheme != 'https'
       super(**args)
     end
 
@@ -17,7 +17,7 @@ module Licensee
     end
 
     def files
-      [{name: @uri.to_s, uri: @uri}]
+      [{ name: @uri.to_s, uri: @uri }]
     end
   end
 end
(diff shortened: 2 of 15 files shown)


lib/licensee/license.rb
```diff
@@ -63,7 +63,7 @@ def licenses
     # `no-license` - The project is not licensed (e.g., all rights reserved)
     #
     # Note: A lack of detected license will be a nil license
-    PSEUDO_LICENSES = %w[other no-license].freeze
+    PSEUDO_LICENSES = %w(other no-license).freeze
```
mlinksva: You're probably using a recent but not current version of rubocop. This was fixed/changed back in https://github.com/bbatsov/rubocop/pull/4231

willbar: Thanks, I'm pretty unfamiliar with Ruby, I deleted my ```Gemfile.lock``` and reinstalled the dependencies and yes that made those lint errors go away, I've reverted those changes.

benbalter:
@willbar Thanks for this. Can you explain a bit more about the problem you're trying to solve? If you have the URL to the file, would `curl URL | licensee` work, or do you need the data in Ruby-land?

The general implementation looks good, although I'm a bit concerned about making remote HTTP calls by default. We use Licensee in production at GitHub.com, and I can imagine some edge cases where an unsanitized or network path might make it's way to Licensee.

How would you imagine this being used? Being pointed at the URL to a text file? An entire Git repo? A GitHub repo? Part of the value of the project is being able to determine the LICENSE file, which may be `LICENSE-MIT`, `COPYING`, etc.

willbar:
@benbalter Sure, when registering open source projects we need to determine the license the project is under, for projects that are in a GitHub repository we do call the GitHub API today to get that information (leveraging all of the goodness you mention today) but in other cases we just have a URL to some random file and we need to be able to determine the license. We need to be able to do this in a service so shelling out isn't ideal.

What if I changed the code to allow you to opt-in that project type?

benbalter:
> What if I changed the code to allow you to opt-in that project type?

Makes sense, and thanks for the additional context.

willbar:
@benbalter What I ended up doing was adding a ```allow_remote``` arg (similar to ```detect_packages```) that defaults to ```false```. The command line (```bin\licensee```) sets it to ```true``` so you can do URL lookups from the command-line. I'm assuming the GitHub API uses the library not the command-line, correct?

The advantage of adding it as an arg is if you add new project types in the future that might do remote calls they can use the arg to determine if that capability is enabled or not.

Does this feel like it's on the right track? I need to add some more unit tests but I wanted to validate the approach with you first.

lib/licensee/projects/uri_project.rb
```diff
@@ -0,0 +1,25 @@
+# URI-based project
+#
+# Analyzes a given URI for license information
+require 'net/http'
+require 'uri'
+
+module Licensee
+  class UriProject < Project
+    def initialize(path, **args)
```
benbalter: I think we'll want to raise an `UnsuportedProeject` error here if `allow_remote` is false, rather than raising when `load_file` is first called. This matches the behavior of FSProject, and will allow us to add an additional project type in the last after URIProject if we want down the line. This will also allow us to get ride of the `@allow_remote` guards on each method.

lib/licensee/projects/uri_project.rb
```diff
@@ -0,0 +1,25 @@
...
+      @uri = URI(path)
```
benbalter: We may need to rescue from an invalid URI error here, in the event that I pass "http://foo"

lib/licensee/projects/uri_project.rb
```diff
@@ -0,0 +1,25 @@
...
+      raise UnsupportedProject if @uri.scheme !~ /^https?$/
+      super(**args)
+    end
+
+    def load_file(file)
```
benbalter: For every other project-type, `load_file` take a hash with key `name`, which is relative to the project root. Perhaps to mirror that behavior, `load_file` should expect the same hash, but URI.join the relative path to the root URI?

lib/licensee/projects/uri_project.rb
```diff
@@ -0,0 +1,25 @@
...
+      Net::HTTP.get(file[:uri]) if @allow_remote
+    end
+
+    def files
```
benbalter: For other project types, `files` would return an array of files within the project. Obviously that's not possible here. What if we just made `@files` an `attr_acessor` and then a user would just do `files.push(name: "LICENSE")` with any files they knew to exist, and LICENSE can support multiple license files, multiple matching strategies, etc.

Alternatively, we could have a hard-coded list of the most common paths, call them in order, and only go to the next one if the previous doesn't exist (e.g., `LICENSE`, `LICENSE.txt`, `LICENSE.md`, etc.).

PR #203, opened 2017-06-22
Multiple license support

benbalter:
This PR abstracts license detection to allow better support for multiple licenses. In terms of public (Ruby) API, the proposed behavior, based on @mlinksva's https://github.com/benbalter/licensee/pull/202 is a follows:

* Project with a single, detectable license file: `license_file` returns the license file and `license` returns the license (current behavior).
* Project with a single, undetectable license file: `license_file` returns the license file and `license` returns nil (current behavior).
* Project with multiple license files and multiple licenses: `license_file` returns nil and `license` returns nil. `licenses_files` returns an array of LicenseFiles and `licenses` returns an array of Licenses.
* Project with multiple license files and a single license: `license_file` returns nil and `license` returns the License. `licenses_files` returns an array of LicenseFiles and `licenses` returns an array with a single License.

Under the hood, we're creating an array of LicenseFiles, and if there's only one, using the first (with an exception for LGPL as before). The logic if further extended that if there are multiple matched files, but all have the same license, it is consider single, not dual licensed.

The LGPL detection is now more strict, in that it requires:

* A license file that's detected as GPL, and
* A `COPYING.lesser` file *that's detected as LGPL*

If both conditions are met, it's considered single licensed, despite having two matched license files.

An upstream consumer would likely check `licenses` to see if the project is single or dual licensed (or just use the `license` helper), and the existence of `license_files` (or the `license_file` helper to see if the project is licensed at all, but undetected. I toyed with the idea of returning `Other` for dual-licensed code, but in the end decided to match the old behavior as closely as possible.

Here's what the command line output might look like:

```
License: MIT License
Matched files: ["LICENSE.md", "licensee.gemspec"]
LICENSE.md:
  Content hash: d64f3bb4282a97b37454b5bb96a8a264a3363dc3
  Attribution: Copyright (c) 2014-2016 Ben Balter
  Confidence: 100.00%
  License: MIT License
licensee.gemspec:
  Confidence: 90.00%
  Matcher: Licensee::Matchers::Gemspec
  License: MIT License
```

One note,  this will likely require a major version bump. In implementing this, which called `uniq?` on license objects, I discovered that we were overwriting the `Object#hash` method with a string, but some native Ruby functions expect this to be an Integer. We could code around this, if we want to maintain backwards compatibility (e.g., require users to pash a block to `uniq`), but to do things the "Right" way, I renamed `hash` to `content_hash` which will be a breaking change.

Thoughts on this approach?

Fixes https://github.com/benbalter/licensee/issues/57, Fixes https://github.com/benbalter/licensee/issues/114.


benbalter:
Here's an example of matching a dual-licensed project ([Rust](https://github.com/rust-lang/rust)):

```
Licenses: ["Apache License 2.0", "MIT License"]
Matched files: ["LICENSE-APACHE", "LICENSE-MIT"]
LICENSE-APACHE:
  Confidence: 100.00%
  License: Apache License 2.0
LICENSE-MIT:
  Content hash: d64f3bb4282a97b37454b5bb96a8a264a3363dc3
  Attribution: Copyright (c) 2010 The Rust Project Developers
  Confidence: 100.00%
  License: MIT License
```

mlinksva:
Seems like a good approach overall.

Couple small things:

> I toyed with the idea of returning Other for dual-licensed code, but in the end decided to match the old behavior as closely as possible.

I [like](https://github.com/benbalter/licensee/pull/190) using `other` as it requires a programmer using the library to do a bit less -- now if they want to distinguish no detected license from no license file, they have to look at both (and with this, also licenses), but it's not a big deal either way. I'd suggest removing the `other` pseudolicense from the code and tests if we're not going to use it.

> Here's what the command line output might look like:
> ```
> License: MIT License
> Matched files: ["LICENSE.md", "licensee.gemspec"]
> LICENSE.md:
>   Content hash: d64f3bb4282a97b37454b5bb96a8a264a3363dc3
>   Attribution: Copyright (c) 2014-2016 Ben Balter
>   Confidence: 100.00%
>   Matcher: Licensee::Matchers::Exact
>   License: MIT License
> licensee.gemspec:
>   Confidence: 90.00%
>   Matcher: Licensee::Matchers::Gemspec
>   License: MIT License
> ```

I like this output, but doesn't it require more than is described (and I think more than is implemented, though I haven't read the code closely) here? I think currently package files are never looked at unless a detected license from a license file is not found. What happens if the license file and package file contain different licenses?

benbalter:
> I like using other as it requires a programmer using the library to do a bit less

That makes sense. If there's a `matched_file` file, we'll always return `license`, even if it's `other`, such that the only time `license` is `nil` is when there's no `matched_files`.

Looking at the code, it's a small change to have `ProjectFile#match` fallback `other`, but since it affects single-licensed projects as well, think it's best tackled in a follow up PR.

> I think currently package files are never looked at unless a detected license from a license file is not found. What happens if the license file and package file contain different licenses?

With this PR, if package file support is enabled, it will look at both license file and package files. If they are the same, there's not a problem. If they differ, it'll be considered a dual licensed project.

mlinksva:
@benbalter :+1: I didn't read the last bullet carefully enough:

> Project with multiple license files and a single license: license_file returns nil and license returns the License. licenses_files returns an array of LicenseFiles and licenses returns an array of Licenses.

Which covers case illustrated by command line output, where `LICENSE` and package metadata agree on single license.

sergey-alekseev:
Is it supposed to fetch multiple licenses from `*.gemspec`?  

E.g. `s.licenses = ["Apache-2.0", "Artistic-2.0", "BSD-3-Clause", "CC-BY-NC-SA-4.0", "AGPL-3.0", "EPL-1.0", "GPL-2.0", "GPL-3.0", "LGPL-3.0", "MIT", "MPL-2.0", "Ruby"]` in https://github.com/SixArm/sixarm_ruby_unaccent/blob/f8f0819dfd433bce8f9e3862b2a8192fa11de476/sixarm_ruby_unaccent.gemspec#L13.  


```
$ licensee detect https://github.com/SixArm/sixarm_ruby_unaccent
License:        NOASSERTION
Matched files:  LICENSE.md, sixarm_ruby_unaccent.gemspec
LICENSE.md:
  Content hash:  3b41f660ac82b6c90a0a8951deb1d7ad7b1dd90c
  License:       NOASSERTION
sixarm_ruby_unaccent.gemspec:
  Confidence:  90.00%
  Matcher:     Licensee::Matchers::Gemspec
  License:     NOASSERTION

$ irb
001 > require 'licensee'
 => true 
002 > license = Licensee.license 'https://github.com/SixArm/sixarm_ruby_unaccent'
 => #<Licensee::License key=other> 
003 > license.key
 => "other" 
```

(routed from https://github.com/github/licensed/issues/189#issuecomment-552973434)

mlinksva:
@sergey-alekseev no licensee can't currently parse more than a single license in a gemspec (or other package manager file). Even if it could, I'm not sure of the semantics of an array of licenses in the example you give or the documentation for that use https://guides.rubygems.org/specification-reference/#licenses= ... a SPDX license expression with `OR` or `AND` would be more clear (but licensee can't currently parse that either).

PR #204, opened 2017-06-26
Return the "other" license when a license is found but cannot be identified

benbalter:
An alternative to https://github.com/benbalter/licensee/pull/190, taking into account to the multiple license support provided in https://github.com/benbalter/licensee/pull/203, and per @mlinksva's comments over in https://github.com/benbalter/licensee/pull/203#issuecomment-310448532, this PR changes Licensee's behavior, such that if we detect the author suspected to license their project, but are unable to identify the license, we return the "other" pseudo license, rather than `nil`, with `nil` responses indicating that no intent to license was detected.

A few examples of what `Project#license` would return:

* `LICENSE` file with an unknown license - "other" license
* Project with no license file - `nil`
* Project with multiple identified licenses - "other" license (`licenses` would return an array of licenses)
* Project in which the `LICENSE` file and package manager license conflict - - "other" license (`licenses` would return an array of licenses)
* Project with a package file (e.g. Gemspec) with no license property - `nil`
* Project with a package file (.e.g, Gemspec) with an unknown license value - "other" license

Thoughts on this behavior? This would be a breaking change and require a major version bump.

In either event, thanks for spiking out the early work in #190 and for pushing the conversation.

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -4,7 +4,8 @@ require 'yaml'
 module Licensee
   class InvalidLicense < ArgumentError; end
   class License
-    @all ||= {}
+    @all = {}
+    @keys_licenses = {}
 
     class << self
       # All license objects defined via Licensee (via choosealicense.com)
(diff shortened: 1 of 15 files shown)


benbalter:
> I wonder if while making a breaking change, we shouldn't introduce a 'multiple' pseudolicense, and return that in the case 'other' would be returned due to finding multiple licenses?

👍 . Easy to implement. Will get that going in a subsequent PR.

PR #205, opened 2017-06-26
Properly normalize CRLF line endings

benbalter:
@mlinksva in testing https://github.com/benbalter/licensee/pull/204, I couldn't understand why the CFPL Git test kept failing for me locally. Checking the output, I realized I have global LR/LF normalization set for Git, the CFPL fixture had Windows line endings, and we only striped unix-style line endings.

This PR fixes the content normalization to also normalize `\r\n` line endings, adds a test, and converts the CFPL fixture to use `\n` line endings.

In doing so, however, the FCPL test now fails (it's detected as MPL). I believe this was a problem with the test, not the implementation, and now accurately captures the real world behavior (meaning we were not and are not properly distinguishing FCPL from MPL).

```
➜  licensee git:(windows-line-endings) be licensee spec/fixtures/fcpl-modified-mpl
License: Mozilla Public License 2.0
Matched files: ["LICENSE"]
LICENSE:
  Content hash: 50413f264978001a3e78c5a1966bea5a7c0fb84b
  Confidence: 98.32%
  Matcher: Licensee::Matchers::Dice
  License: Mozilla Public License 2.0
  Closest licenses:
    * MPL-2.0 similarity: 98.32%
```


mlinksva:
Hmm, not sure how I managed to check in a file with CRLF endings, sorry about that. I'll take a closer look at properly distinguishing FCPL from MPL, hopefully in the next day.

benbalter:
@mlinksva no need to apologize. Had me puzzled since it merged. 

I marked the FCPL test as skipped via a643d55, will get this PR merged with the skipped but failing test so that you can more easily play with the delta/matching on master in a subsequent PR.

PR #206, opened 2017-06-26
make test match updated (c) year

mlinksva:


benbalter:
My bad. Assumed it was a simple change. Thanks for cleaning things up.

PR #209, opened 2017-06-29
enable file system license searching in parent directories

jonabc:
This adds an optional `:search_root` argument to the FsProject initializer.  When set, it enables `FsProject` to search all directories between `path` and `:search_root` for license files.

This is not a breaking change.  If `:search_root` is not set, the previous behavior is used and `FsProject#files` will not search the directory hierarchy.


jonabc:
Needed to add full metadata storage into `project_file.rb` so that users of the library can construct a full path to files if they're found outside of the initial search path.

There are quite a few methods in `project.rb` that call `files` multiple times.  During a typical run I think I saw `files` being called 5+ times per dependency.  It would make sense to cache the searched files.  It's outside the scope of this change but happy to add it if wanted.  💭 ?

mlinksva:
I've noticed the repetitive calls as well and 🤔 for a moment each time. But changing this implementation detail wasn't necessary and I didn't know whether it's making the right tradeoff. So I'm idly curious about what the right thing to do is.

jonabc:
@benbalter how do you feel about this change?  This change, or something like it, will be needed to support `FsProject` with Go, where `import`ed packages are often subdirectories of an overall project.

lib/licensee/projects/fs_project.rb
```diff
@@ -19,11 +20,16 @@ def initialize(path, **args)
     # Returns an array of hashes representing the project's files.
     # Hashes will have the :name key, with the relative path to the file
     def files
+      dir = @dir
       files = []
 
-      Dir.glob(::File.join(@dir, @pattern).tr('\\', '/')) do |file|
-        next unless ::File.file?(file)
-        files.push(name: ::File.basename(file))
+      while search_directory?(dir)
```
benbalter: If we rely on Pathname, could this be simplified to:

```ruby
root = Pathname.new(@root)
Pathname.new(dir).relative_path_from(root).each
...
```
?

benbalter:
>  It would make sense to cache the searched files. It's outside the scope of this change but happy to add it if wanted. 

You're talking about memoizing https://github.com/benbalter/licensee/blob/125d93fd347a5148040f2da93b32e51ecd977084/lib/licensee/projects/git_project.rb#L49-L59 and https://github.com/benbalter/licensee/blob/125d93fd347a5148040f2da93b32e51ecd977084/lib/licensee/projects/fs_project.rb#L19-L30? If so 👍. Can do it here, or if we wanted to do it "right", glad to merge a PR that just memoizes those two methods (if that's what you're suggesting).

jonabc:
@benbalter I think I've updated things to be a little bit more understandable and make better use of `Pathname`.

I included project file memoizations for each of the project types since it didn't add much to the PR

benbalter:
@jonabc This is great. Thanks for taking the time to work through the feedback. I agree this is much more readable (and makes me less nervous about file system access if we're using core libraries). Have small small tidying up I want to do to how some of the other classes are named that I was waiting for this on, and will get this out in the next major release. Thanks!

PR #210, opened 2017-06-29
Cabal project file matcher

jonabc:
Adds matching for Haskells cabal files which are yaml.

**NOTE:** Also fixed a small issue with the npm/bower matcher regex.


benbalter:
This is great! Thanks @jonabc!

PR #212, opened 2017-06-30
Fix for FCPL false positive

benbalter:
@mlinksva This is a different approach to solve for the FCPL false positive. Rather than hard-coding limits like the max delta, this PR:

1. Increases the confidence threshold to 98%, which makes the max delta smaller across all licenses
2. Better content normalization so the delta is more accurate (whitespace and HRs)

Beyond that this PR also fixes the integration test that modified licenses - previously it added a small number of words to the end of the license which (A) was a small change when converted to a wordset and (B) was stripped entirely from licenses with instructions. Now modified licenses should not match, which I believe is the intended behavior (we were previously testing that they *were* matching).

Raising the confidence threshold is obviously going to result in more false negatives for modified licenses, but (A) I think that's the intended behavior in many cases (like FCPL) and (B) I'd prefer we return false negatives to false positives.


mlinksva:
❤️ : 

- better normalization
- higher confidence threshold
- removal of 3 random added words expected to pass tests (a mild source of aggravation as an occasional contributor)

The only concern I have is that I guess that false negative rate will be much higher for old short permissive licenses, as their bodies have historically been templated. That was the hunch I was following in trying to put hard limits on the changed words/length for longer licenses, while keeping 5% wiggle room for shorter licenses.

I strongly agree on preferring false negatives to false positives and totally fine with choices made in this PR, just putting this thought out there in case useful.

benbalter:
:metal: 

> The only concern I have is that I guess that false negative rate will be much higher for old short permissive licenses, as their bodies have historically been templated. That was the hunch I was following in trying to put hard limits on the changed words/length for longer licenses, while keeping 5% wiggle room for shorter licenses.

I thought about suggesting a minimum delta, but anything more than 2% is too much to distinguish FCPL from MPL.

Issue #196, opened 2017-05-26
Handling custom licenses


hyandell:
I've been playing with monkeypatching an additional custom license location in. This seems to be working for me:
```
Licensee::License.class_eval do

    def self.add_custom_license_dir(dir)
      @custom_license_dir=dir
    end

    def self.custom_license_dir
      return @custom_license_dir
    end

    def self.license_files
      dir=license_dir()
      files ||= Dir.glob("#{license_dir()}/*.txt")
      files.concat(Dir.glob("#{custom_license_dir()}/*.txt"))
      return files
    end

    def path
       path ||= File.expand_path "#{@key}.txt", Licensee::License.license_dir
       unless(File.exist?(path))
         path = File.expand_path "#{@key}.txt", Licensee::License.custom_license_dir
       end
       return path
    end

end
```
Would you be interested in a patch that makes that more part of the API? I would probably make custom_license_dir an array rather than a single element. Given it has to spend time doing File.exist in path, it could go further and change the path API to return nil if the files don't exist.

hyandell:
Noting that I've added it to oss-dashboard:
  https://github.com/amzn/oss-dashboard/commit/e744c8c3015b84e8999a225ec87c0a4212de2e90
  https://github.com/amzn/oss-dashboard/commit/6b7896e7248e198515503f8c7288103f8b356450

benbalter:
@hyandell can you explain a bit more about what you're trying to do here?

hyandell:
Sure (sorry for the delay).

I'm using licensee on my own GitHub repositories. I want to support more licenses than the choosealicense or spdx repositories have, so I have a 2nd directory that contains my customer license files (in the same format as choosealicense/licensee).

The monkeypatching above splices that extra directory in so that licensee is looking at multiple paths for canonical licenses. Ideally I'd get that into the API as I'm sure the monkeypatching will bite me :)

mlinksva:
I can see this being useful. It might be good to have an additional flag which directs whether licenses in the custom directory override the built-in ones. In the monkeypatch above, the custom licenses are just added to the built-in list, but the path returned is always the built-in one. I'm not sure what behavior would result in the case of a duplicate.

A use case for overriding might be some kind of self-testing in choosealicense.com, where licensee vendors its licenses from. Maybe I'll try writing a monkeypatch version over there to clarify.

hyandell:
[...]

PR #216, opened 2017-07-25
Adding license_dirs function allowing you to add other license locations

hyandell:
First attempt at issue #196 

Let me know what you think. 

  Licensee::License.license_dirs=[my_special_dir, Licensee::License.default_license_dir]
  project=Licensee::Projects::GitProject.new(review_directory)

It's only useful at the code level right now. Next step would be to add an environment variable or command line argument.

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -8,6 +8,18 @@ module Licensee
     @keys_licenses = {}
 
     class << self
+      def default_license_dir
+        dir = ::File.dirname(__FILE__)
+        ::File.expand_path '../../vendor/choosealicense.com/_licenses', dir
+      end
+
+      # Backwards compatability
+      alias license_dir default_license_dir
+    end
+
+    @license_dirs = [ default_license_dir() ]
+
+    class << self
       # All license objects defined via Licensee (via choosealicense.com)
       #
       # Options:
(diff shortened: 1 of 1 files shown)


lib/licensee/license.rb
```diff
@@ -38,13 +50,23 @@ def find(key, options = {})
       alias [] find
       alias find_by_key find
 
-      def license_dir
-        dir = ::File.dirname(__FILE__)
-        ::File.expand_path '../../vendor/choosealicense.com/_licenses', dir
+      def license_dirs
+        return @license_dirs
```
benbalter: Why not make this:

```ruby
def license_dirs
  @license_dirs ||= File.expand_path '../../vendor/choosealicense.com/_licenses', File.dirname(__FILE__)
end
```

lib/licensee/license.rb
```diff
@@ -38,13 +50,23 @@ def find(key, options = {})
...
+      end
+
+      # A user customized lookup array of directories for license files. If the user wants 
+      # to include the default, they should make one of the entries be default_license_dir()
+      def license_dirs=(dirs)
+        @license_dirs=dirs
```
benbalter: I believe this can be `attr_writer :license_dirs`

lib/licensee/license.rb
```diff
@@ -38,13 +50,23 @@ def find(key, options = {})
       alias [] find
       alias find_by_key find
 
-      def license_dir
```
benbalter: For backwards compatibility, we'll want a class method 

```ruby
def license_dir
  license_dirs.first
end
```

That way, it remains accurate even if overwritten.

lib/licensee/license.rb
```diff
@@ -38,13 +50,23 @@ def find(key, options = {})
...
       end
 
+      # Returns all the license files across the various license directories
       def license_files
-        @license_files ||= Dir.glob("#{license_dir}/*.txt")
+        files=[]
+        @license_dirs.each do |dir|
+          files.concat(Dir.glob("#{dir}/*.txt"))
+        end
+        @license_files ||= files
```
benbalter: This doesn't properly memoize the method because the `Dir.glob` logic will always run.  How about

```ruby
@license_files ||= license_dirs.map { |d| Dir.glob("#{d/*.txt") }.flatten
```

lib/licensee/license.rb
```diff
@@ -82,7 +104,14 @@ def initialize(key)
...
-      @path ||= File.expand_path "#{@key}.txt", Licensee::License.license_dir
+      foundpath=nil
+      Licensee::License.license_dirs().each do |dir|
+        foundpath = File.expand_path "#{@key}.txt", dir
+        if(File.exist?(foundpath))
+          break
+        end
+      end
+      @path = foundpath
     end
```
benbalter: Not a huge fan of having to search for file existence across multiple directories. Is there an alternate way we could do this?

PR #231, opened 2017-10-25
Support for square brackets in license text

benbalter:
Upstream change to corespond with https://github.com/github/choosealicense.com/pull/547 and https://github.com/github/choosealicense.com/issues/542.

This adds a `Licence#fields` method which returns an array of LicenseField objects, as well as a `content_for_mustache` method which returns the mustache-prepared content, aware of the supported fields (and ignoring all others).

One gotcha... in fixing how we stub in copyright info during tests, I discovered the Postgres license isn't being properly detected in our tests due to the fact that it has the `[fullname]` field multiple times (previously we were only replacing the first instance). That can be fixed in another pass as it's technically an unrelated change.

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -143,6 +143,19 @@ module Licensee
       "#<Licensee::License key=#{key}>"
     end
 
+    # Returns an array of strings of substitutable fields in the license body
+    def fields
+      @fields ||= LicenseField.from_content(content)
+    end
+
+    # Returns a string with `[fields]` replaced by `{{{fields}}}`
+    # Does not mangle non-supported fields in the form of `[field]`
+    def content_for_mustache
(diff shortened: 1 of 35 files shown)


lib/licensee/license.rb
```diff
@@ -143,6 +143,19 @@ def inspect
       "#<Licensee::License key=#{key}>"
     end
 
+    # Returns an array of strings of substitutable fields in the license body
+    def fields
+      @fields ||= LicenseField.from_content(content)
+    end
+
+    # Returns a string with `[fields]` replaced by `{{{fields}}}`
+    # Does not mangle non-supported fields in the form of `[field]`
```
mlinksva: I see how you've accomplished this by building a regex matching any of the fields specified in choosealicense.com's field metadata, in `license_field.rb`. :party:

PR #232, opened 2017-10-28
support LICENSE_ files

### LINKED ISSUE #221: Please look for filename LICENSE_1_0.txt
Projects using the Boost Software License (http://www.boost.org/LICENSE_1_0.txt) usually store their license in a file named `LICENSE_1_0.txt`. The BSL is a license which is widely used by the C++ community (https://github.com/search?utf8=%E2%9C%93&q=Boost+filename%3ALICENSE_1_0.txt&ref=simplesearch).

Would it be possible to add this file name to the list of recognized options for github to detect?

Thanks!

mlinksva:
Fixes #221


lib/licensee/project_files/license_file.rb
```diff
@@ -34,11 +34,12 @@ class LicenseFile < Licensee::ProjectFiles::ProjectFile
         /#{COPYING_REGEX}-/                          => 0.65,  # COPYING-MIT
         /-#{LICENSE_REGEX}/                          => 0.60,  # MIT-LICENSE-MIT
         /-#{COPYING_REGEX}/                          => 0.55,  # MIT-COPYING
-        /\A#{OFL_REGEX}#{PREFERRED_EXT_REGEX}/       => 0.50,  # OFL.md
-        /\A#{OFL_REGEX}#{ANY_EXT_REGEX}/             => 0.45,  # OFL.textile
-        /\A#{OFL_REGEX}\z/                           => 0.40,  # OFL
-        /\A#{PATENTS_REGEX}\z/                       => 0.35,  # PATENTS
-        /\A#{PATENTS_REGEX}#{ANY_EXT_REGEX}\z/       => 0.30,  # PATENTS.txt
+        /#{LICENSE_REGEX}_/                          => 0.50,  # LICENSE_1_0.txt
```
mlinksva: Another approach would be to just accept _ wherever - is. I don't think filename scoring is important now that we support multiple license files (but I could be missing something), so a new score for a new pattern isn't important.

Another approach would be to add a match for LICENSE_1_0.txt ie exactly the filename mentioned in #221.

spec/integration_spec.rb
```diff
@@ -148,6 +148,15 @@
             expect(subject.license).to eql(license)
           end
         end
+
+        context 'BSL' do
```
mlinksva: The integration test and corresponding fixture aren't really necessary because this PR really just changes license file matching, which is tested for in `license_file_spec.rb`, but it doesn't hurt either.

lib/licensee/project_files/license_file.rb
benbalter: Very true. We still use it within GitHub.com. Even if the license is ultimately deemed `other`, we still link to the preferred license file. This could potentially be a breaking change we queue up for the next major version.

lib/licensee/project_files/license_file.rb
```diff
@@ -34,11 +34,12 @@ class LicenseFile < Licensee::ProjectFiles::ProjectFile
         /#{COPYING_REGEX}-/                          => 0.65,  # COPYING-MIT
```
benbalter: What if instead, we changed the `-` on lines 33 and 34 to `[\-_] and just made the existing regex slightly more flexible?

benbalter: Also to note, I believe the regexes in lines 33 and 34 should be preceded by a `\A` to ensure it starts the file name.

lib/licensee/project_files/license_file.rb
```diff
@@ -34,11 +34,12 @@ class LicenseFile < Licensee::ProjectFiles::ProjectFile
...
         /-#{LICENSE_REGEX}/                          => 0.60,  # MIT-LICENSE-MIT
         /-#{COPYING_REGEX}/                          => 0.55,  # MIT-COPYING
```
benbalter: Likewise, I think we can make these `[\-_]`.

lib/licensee/project_files/license_file.rb
mlinksva: Does GitHub.com link to a preferred license file even if `other` is detected? I can't think of where.

Issue #230, opened 2017-10-24
COPYRIGHT file confuses github


purpleidea:
Hi,

After running a re-license (seen here: https://github.com/purpleidea/mgmt/commit/46be83f8f79c8cb81fdf1eae9d951c70c69246b4) on my code base, GitHub no longer detects the license correctly. GH logs seem to indicate it is getting confused by the "copyright" file.

I'm assuming this is a regression, and I figured I'd ask here before I dug into the code. In particular my repo is in a very standard and common format, so I would expect this to work.

Thanks,
James

mlinksva:
It's intentional, see #202.

purpleidea:
@mlinksva Aha, thanks!

So the bug is that since the two files are compatible, this is breaking. I think the more correct logic should be: "of places where we've detected a license", see if there are ambiguities.

This would solve both these issues.

mlinksva:
Yes that'd be ideal, but it'd mean we'd have to accurately detect license notices like the one in https://github.com/purpleidea/mgmt/commit/46be83f8f79c8cb81fdf1eae9d951c70c69246b4#diff-97d9303acdfc078a050e61dc5c1a9a76

It's possible but there's no code at all to do that. There is a catalog of notices to match against in SPDX's standard license header, eg https://spdx.org/licenses/GPL-3.0#licenseHeader

akien-mga:
Just FYI, this broke license detection for Godot too: https://github.com/godotengine/godot/pull/12531
Is there a list of "safe names" that won't be parsed by licensee?

purpleidea:
I think at this point a file in .github/legal with some sort of simple format like:
```
ignore: COPYRIGHT
ignore: foo/*.sh
```

or

```
license: GPL3+
```

in it would probably be a good approach. (I'd send a patch but GH itself is closed source!) :P

mlinksva:
@akien-mga see https://github.com/benbalter/licensee/blob/3a2015eeedbb7d21603b132e9f3bf55a4f0237a8/lib/licensee/project_files/license_file.rb#L6-L43 for definition of filenames that licensee looks at.

Issue #222, opened 2017-08-21
What should licensee do about an inconsistent license identifier in the README when more definitive license indicators are present?


benbalter:
### Problem

 In [Primer's README](https://github.com/primer/primer-css/tree/80a2eca00b19ba0392d7f5d0955d784eae632725/modules/primer-css#license), it has a `license` heading, with `MIT © GitHub`. 

Prior to v9.x, Licensee had no concept of multiple licenses, and would prefer the LicenseFile over the PackageManagerFile over the ReadmeFIle license section. 

With 9.x, Licensee now identifies each individually, rather than stopping with the first. In the case of Primer, it gets MIT from the license file, MIT from the package manager, and "other" from the README, which means it's a mixed license and returns "other" for `Project#license`.

To note, if the README had the full license text, all three sources would report "MIT", meaning it'd be MIT licensed.

### Solutions

1. Ignore the README heading if other, more definitive sources exist (or put another way, only look at the README if no other sources exist, or only look at the README if its license intent is conclusive). This would work fine for primer, but what if the README said "GPLv3" even though there was MIT license text in the license file or a MIT designation via the package manager. Would a downstream user want Licensee to flag this inconsistency? (I think so).

2. Some sort of fuzzy heuristics on matching the content of the license section of the README with the title / nickname of the suspected license.

3. Continue the current behavior and have downstream consumer migrate to `Project#license`**s** and let them make their own determination / flag it for the user etc. (cc @bkeepers, @jonabc)

4. Drop README support entirely.

Thoughts?

shawnbot:
I would suggest ignoring the README if you get a valid license from the license file or the package manager file (in that order), especially if you get the same one from both. Personally, I've never written a README that references the license because GitHub and npm both do that for me.

mlinksva:
:+1: to ignoring `README` if license detected in license file or package manager file.

I think looking for potentially conflicting license info maybe should be considered as a separate feature. Could be in README, could be elsewhere in project, e.g., source code comments.

benbalter:
Fixed via https://github.com/benbalter/licensee/pull/233.

PR #233, opened 2017-11-03
Better handling of false 'others' due to multi-license support

benbalter:
This PR attempts to tackle #230 and #222 in related but distinct ways:

1. If a `COPYRIGHT` file is present and contains only copyright notice (the existing copyright matcher), don't count it as a second license for purposes of determining dual licensing.

2. If a consumer opts in into checking the README, and the README contains a *reference* to a known license (title, nickname, or key), consider the README as indicating license to prevent cases when a LICENSE file says one licence and the README contains a license heading, but not the license text, causing a false `other`. This is implemented as the `Reference` matcher, available only to README files.

To accomplish the above I added:

1. A `License#find_by_title` method, which matches on title, nickname, or key.
2. Added a `pseudo` argument to `License#all` to exclude `other` and `no-license`

One gotcha to call out:

To keep the logic in `Project` simple, I opted to add the new logic as a matcher. That means, that if a consumer opts into README file detection, it'll be naive, in that if a README has a license section with `The MIT License`, and nothing else (and no license file), the project as a whole will be considered MIT. 

In #222 we talked about only employing that strategy when the license is detected by other means, and the README license is consistent with that other license. That's possible to implement, but would add some complexity, so I wanted to pause and check in before I went any further.

Fixes https://github.com/benbalter/licensee/issues/230. Fixes https://github.com/benbalter/licensee/issues/222.

### DIFF
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -6,11 +6,7 @@ module Licensee
     DIGEST = Digest::SHA1
     END_OF_TERMS_REGEX = /^\s*end of terms and conditions\s*$/i
     HR_REGEX = /[=\-\*][=\-\*\s]{3,}/
-    ALT_TITLE_REGEX = {
-      'bsd-2-clause'       => /bsd 2-clause( \"simplified\")? license/i,
-      'bsd-3-clause'       => /bsd 3-clause( \"new\" or \"revised\")? license/i,
-      'bsd-3-clause-clear' => /(clear bsd|bsd 3-clause( clear)?) license/i
-    }.freeze
+    ALT_TITLE_REGEX = License::ALT_TITLE_REGEX
     ALL_RIGHTS_RESERVED_REGEX = /\Aall rights reserved\.?$/i
     WHITESPACE_REGEX = /\s+/
     MARKDOWN_HEADING_REGEX = /\A\s*#+/
@@ -62,7 +58,7 @@ module Licensee
         string = content.strip
         string = strip_markdown_headings(string)
         string = strip_hrs(string)
-        string = strip_title(string) while string =~ title_regex
+        string = strip_title(string) while string =~ ContentHelper.title_regex
         strip_version(string).strip
       end
     end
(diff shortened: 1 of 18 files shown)


lib/licensee/content_helper.rb
```diff
@@ -112,21 +108,25 @@ def self.format_percent(float)
...
+    def self.title_regex
+      licenses = Licensee::License.all(hidden: true, psuedo: false)
+      titles = licenses.map(&:title_regex)
 
-    def license_names
-      @license_titles ||= License.all(hidden: true).map do |license|
-        regex = ALT_TITLE_REGEX[license.key]
-        regex || license.name_without_version.downcase.sub('*', 'u')
+      # Title regex must include the version to support matching within
+      # families, but for sake of normilization, we can be less strict
```
mlinksva: norm**a**lization

lib/licensee/license.rb
```diff
@@ -70,6 +78,19 @@ def keys_licenses(options = {})
...
+    DEFAULT_OPTIONS = {
+      hidden:   false,
+      featured: nil,
+      psuedo:   true
+    }.freeze
+
+    ALT_TITLE_REGEX = {
+      'bsd-2-clause'       => /bsd 2-clause(?: \"simplified\")?/i,
+      'bsd-3-clause'       => /bsd 3-clause(?: \"new\" or \"revised\")?/i,
+      'bsd-3-clause-clear' => /(?:clear bsd|bsd 3-clause(?: clear)?)/i
```
mlinksva: Not related to the purpose of this PR (this regex definition is just moved), but just asking while noticing -- the purpose of this is to catch title matches that don't include bits that aren't needed to disambiguate, or also in the case of "clear bsd", an alternate name, right?

If so does it make sense to also add entries for the A/L/GPL licenses matching titles without the "GNU " prefix?

mlinksva:
> In #222 we talked about only employing that strategy when the license is detected by other means, and the README license is consistent with that other license. That's possible to implement, but would add some complexity, so I wanted to pause and check in before I went any further.

I'm alright with not implementing this. It's consistent with how we treat license assertions in package manager files.

lib/licensee/license.rb
benbalter: >  or also in the case of "clear bsd", an alternate name, right?

Exactly. This allows `clear BSD` to match `bsd-3-clause-clear` since there is no ambiguity and some license templates had the alternate wording (for whatever reason).

> If so does it make sense to also add entries for the A/L/GPL licenses matching titles without the "GNU " prefix?

Yes. I can do that here very easily with this refactor. We can make `GNU` optional in all cases.

benbalter: > I can do that here very easily with this refactor. We can make GNU optional in all cases.

Implemented via 25e7388.

benbalter:
Confirmed the original case in #222 is now properly MIT:

```
➜  licensee git:(multi-license-better) ✗ script/git-repo https://github.com/primer/primer-css/
License: MIT License
Matched files: ["LICENSE", "README.md"]
LICENSE:
  Content hash: d64f3bb4282a97b37454b5bb96a8a264a3363dc3
  Attribution: Copyright (c) 2016 GitHub Inc.
  Confidence: 100.00%
  License: MIT License
README.md:
  Content hash: 325c512cefceae05b2e52bd9b9db61f9fb061020
  Confidence: 90.00%
  License: MIT License
```

uldall:
I tried this on the Godot project which have both a LICENSE file and a COPYRIGHT file:
```
$ licensee
License: Other
Matched files: ["LICENSE.txt", "COPYRIGHT.txt"]
LICENSE.txt:
  Content hash: d64f3bb4282a97b37454b5bb96a8a264a3363dc3
  Attribution: Copyright (c) 2007-2017 Juan Linietsky, Ariel Manzur.
  Confidence: 100.00%
  License: MIT License
COPYRIGHT.txt:
  Content hash: 54ab10b5a44f1816030ddfa4cd2b6ff9af77d2ed
  License: Other
```

It doesn't seem to resolve #230.

benbalter:
@uldall that is the expected behavior. The COPYRIGHT file is ignored if it only contains a copyright notice (and no other legally significant text, such as license text). If you look at https://github.com/godotengine/godot/blob/master/COPYRIGHT.txt, it contains a lot of legally significant text which a robot isn't smart enough to parse, hense the "other" designtation.

PR #235, opened 2017-11-10
Match "Apache License, Version 2.0" as Apache-2.0 name

### LINKED ISSUE #234: Apache License 2.0 isn't detected when applied according to their instructions
The [Apache License, Version 2.0](https://www.apache.org/licenses/LICENSE-2.0) contains an appendix, that contains a boilerplate notice that contains the following line: 

```
Licensed under the Apache License, Version 2.0 (the "License"); ...
```

The GitHub repository [wiremod/wire](https://github.com/wiremod/wire) contains the text of the Apache license in the file `license`, and the boilerplate notice verbatim in `readme.md`. This results in the following:

```
$ licensee
License: Other
Matched files: ["license", "readme.md"]
license:
  Confidence: 100.00%
  License: Apache License 2.0
readme.md:
  Content hash: a953ac3d253aaf14bf9a2d62f780811dd8b4f26e
  Attribution: Copyright 2009-2017 Wire Team
  License: Other
```

The issue here is that `readme.md` doesn't mention the phrase "Apache License 2.0", only "Apache License, Version 2.0" (which is the proper name as specified by the Apache License itself). When "Apache License, Version 2.0" is changed to "Apache License 2.0" the correct result is returned:

```
$ sed -i 's/Apache License, Version 2.0/Apache License 2.0/' readme.md
$ git add readme.md && git commit -m 'temp' --quiet
$ licensee
License: Apache License 2.0
Matched files: ["license", "readme.md"]
license:
  Confidence: 100.00%
  License: Apache License 2.0
readme.md:
  Content hash: f4f083f8593f4451b48836075245d53ee12e2976
  Attribution: Copyright 2009-2017 Wire Team
  Confidence: 90.00%
  License: Apache License 2.0
```

I think it makes sense to consider this a bug in Licensee (or choosealicense.com's vendor files), as applying the license using Apache's own boilerplate causes the license to be incorrectly detected.

```
$ gem list licensee

*** LOCAL GEMS ***

licensee (9.5.0)
```

mlinksva:
Apache-2.0's SPDX name is "Apache License 2.0" but the boilerplate at the end of the license text intended for source comments but sometimes used in README notices says "Apache License, Version 2.0". This PR adds recognition of the latter, just as some common unambiguous variations on some other licenses' names are recognized.

Fixes #234

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -86,6 +86,7 @@ module Licensee
     }.freeze
 
     ALT_TITLE_REGEX = {
+      'apache-2.0'         => /apache license(?:, version)? 2.0/i,
       'bsd-2-clause'       => /bsd 2-clause(?: \"simplified\")?/i,
       'bsd-3-clause'       => /bsd 3-clause(?: \"new\" or \"revised\")?/i,
       'bsd-3-clause-clear' => /(?:clear bsd|bsd 3-clause(?: clear)?)/i
--- a/spec/fixtures/apache-with-readme-notice/LICENSE
+++ b/spec/fixtures/apache-with-readme-notice/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
+      "License" shall mean the terms and conditions for use, reproduction,
+      and distribution as defined by Sections 1 through 9 of this document.
+
+      "Licensor" shall mean the copyright owner or entity authorized by
+      the copyright owner that is granting the License.
+
+      "Legal Entity" shall mean the union of the acting entity and all
+      other entities that control, are controlled by, or are under common
+      control with that entity. For the purposes of this definition,
+      "control" means (i) the power, direct or indirect, to cause the
+      direction or management of such entity, whether by contract or
+      otherwise, or (ii) ownership of fifty percent (50%) or more of the
+      outstanding shares, or (iii) beneficial ownership of such entity.
+
+      "You" (or "Your") shall mean an individual or Legal Entity
+      exercising permissions granted by this License.
+
+      "Source" form shall mean the preferred form for making modifications,
+      including but not limited to software source code, documentation
+      source, and configuration files.
+
+      "Object" form shall mean any form resulting from mechanical
+      transformation or translation of a Source form, including but
+      not limited to compiled object code, generated documentation,
+      and conversions to other media types.
+
+      "Work" shall mean the work of authorship, whether in Source or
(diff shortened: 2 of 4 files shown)


benbalter:
I think this makes sense as a good fix for the Apache license. 

More generally, I'm wondering if our license name logic should allow any `[number].[number]` at the end of a license to be preceded by `v` or `version ` (or `, version `). Here's the list of license titles. While not exactly right in SPDX terms, any of the above variations would still convey the same meaning in terms of README references:

```
SIL Open Font License 1.1
GNU Lesser General Public License v3.0
GNU Affero General Public License v3.0
GNU General Public License v2.0
Creative Commons Attribution 4.0
Microsoft Reciprocal License
Do What The F*ck You Want To Public License
Open Software License 3.0
Boost Software License 1.0
GNU Lesser General Public License v2.1
Mozilla Public License 2.0
ISC License
Creative Commons Zero v1.0 Universal
BSD 3-clause Clear License
GNU General Public License v3.0
The Unlicense
BSD 2-clause "Simplified" License
Artistic License 2.0
zlib License
LaTeX Project Public License v1.3c
MIT License
PostgreSQL License
Academic Free License v3.0
University of Illinois/NCSA Open Source License
Creative Commons Attribution Share Alike 4.0
BSD 3-clause "New" or "Revised" License
Eclipse Public License 1.0
Microsoft Public License
Educational Community License v2.0
Apache License 2.0
European Union Public License 1.1
```

mlinksva:
All license title variations would still unambiguous, so acceptable.

Plus: catch some probably small number of license references fitting general pattern that we wouldn't otherwise.

Minus: bow to variation in licensing practice.

Overall, seems reasonable to me. I'll go ahead and merge this to fix #234 in the meantime.

wking:
On Mon, Nov 13, 2017 at 04:29:01PM +0000, Ben Balter wrote:
> While not exactly right in SPDX terms…

I'm not sure what you're concerned about here, but the SPDX matching
guidelines currently recommend completely ignoring the license
name/title when matching license text [1].

That's different from README-matching, where the closest analog is
SPDX's standard copyright header.  The SPDX is trying to just pass
along the upstream template there, and that upstream template
currently includes “Apache License, Version 2.0” [2].

Anyhow, this change looks fine to me.  I was just pointing out that,
as far as I can tell, the README matching *is* exactly right in SPDX
terms.

[1]: https://spdx.org/spdx-license-list/matching-guidelines
     Guideline 11: “Ignore the license name or title for matching
     purposes, so long as what ignored is the title only and there is
     no additional substantive text added here.”
[2]: https://github.com/spdx/license-list-XML/blob/9f4432fbb660510859417b3d78a795beeeb8279b/src/Apache-2.0.xml#L9-L10

benbalter:
@wking Yes. When matching license text, we strip the title and match the license body. For README / reference matching, it's opt-in, and a best guess as to what the author's intent might be.

PR #237, opened 2017-11-13
Allow version notation to be more lenient

benbalter:
As discussed in https://github.com/benbalter/licensee/pull/235#issuecomment-343974853, this PR updates the license title regex to be agnostic to the following version notation formats:

* `The foo license 1.0`
* `The foo license v1.0`
* `The foo license version 1.0`
* `The foo license, version 1.0`
* `The foo license, v1.0`

The idea being, that in all of the above formats, the author's intent is unambiguous, and thus our regex should accept each variation as the matching license.

### DIFF
--- a/lib/licensee/license.rb
+++ b/lib/licensee/license.rb
@@ -86,7 +86,6 @@ module Licensee
     }.freeze
 
     ALT_TITLE_REGEX = {
-      'apache-2.0'         => /apache license(?:, version)? 2.0/i,
       'bsd-2-clause'       => /bsd 2-clause(?: \"simplified\")?/i,
       'bsd-3-clause'       => /bsd 3-clause(?: \"new\" or \"revised\")?/i,
       'bsd-3-clause-clear' => /(?:clear bsd|bsd 3-clause(?: clear)?)/i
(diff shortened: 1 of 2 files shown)


benbalter:
Confirmed that this works with the wire repo @AbigailBuccaneer mentioned over in https://github.com/benbalter/licensee/issues/234#issue-272893126:

```
➜  licensee git:(version-matching-regex) script/git-repo https://github.com/wiremod/wire --license apache-2.0
License: Apache License 2.0
Matched files: ["license", "readme.md"]
license:
  Confidence: 100.00%
  License: Apache License 2.0
readme.md:
  Content hash: a953ac3d253aaf14bf9a2d62f780811dd8b4f26e
  Attribution: Copyright 2009-2017 Wire Team
  Confidence: 90.00%
  License: Apache License 2.0
Comparing to Apache License 2.0:
  Input length: 9060
  License length: 9060
  Similarity: 100.00%
  Exact match!
```

benbalter:
Would be more appropriate as a distinct PR, but if we really want to expand the reference detection, we could also look for references to the license source, since we have it.

PR #238, opened 2017-11-13
Allow ReferenceMatcher to match references to licenses by their source

benbalter:
As suggested in https://github.com/benbalter/licensee/pull/237#issuecomment-344056185, this PR updates ReferenceMatcher to match references to licenses by their source (as defined by choosealicense.com).

To allow for human error, the following permutations are ignored/considered to be the same intent:

* `http` or `https`
* `www.` and non-www
* `.txt`, `.html`, `.htm`, or `/` suffix (or no suffix)

Thoughts? This would only be used if the ReferenceMatcher is used, right now if a downstream users opts into readme detection support.

### DIFF
--- a/spec/licensee/license_spec.rb
+++ b/spec/licensee/license_spec.rb
@@ -387,4 +387,41 @@ RSpec.describe Licensee::License do
       end
     end
   end
+
+  context 'source regex' do
+    Licensee::License.all(hidden: true, psuedo: false).each do |license|
+      context "the #{license.title} license" do
+        let(:source) { URI.parse(license.source) }
+
+        %w[http https].each do |scheme|
+          context "with a #{scheme}:// scheme" do
+            before { source.scheme = scheme }
+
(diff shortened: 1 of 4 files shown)


spec/licensee/license_spec.rb
```diff
@@ -421,4 +421,41 @@
...
+            ['www.', ''].each do |prefix|
+              context "with '#{prefix}' before the host" do
+                before do
+                  source.host = "#{prefix}#{source.host.sub(/\Awww\./, '')}"
+                end
+
+                ['.html', '.htm', '.txt', ''].each do |suffix|
+                  context "with '#{suffix}' after the path" do
+                    before do
+                      next if license.key == 'wtfpl'
```
mlinksva: Is this because the source URL listed for WTFPL doesn't include the license text?

Seems that should be fixed in choosealicense.com, probably http://www.wtfpl.net/txt/copying/ per https://spdx.org/licenses/WTFPL.html

PR #240, opened 2017-11-22
license expression in package.json obtains other rather than nil

mlinksva:
https://docs.npmjs.com/files/package.json#license can be a SPDX
license expression rather than a SPDX license id. If expression
found, makes sense to return other license rather than nil from
package. So add space, plus, and parens to character class match
as those can be part of a license expression.

### DIFF
--- a/spec/fixtures/npm-license-expression/package.json
+++ b/spec/fixtures/npm-license-expression/package.json
@@ -1,0 +1,6 @@
+{
+  "name": "fake-package"
+  "description": "Fake package available under various licenses",
+  "license": "(MIT OR Apache-2.0 OR AGPL-3.0+)",
+  "version": "0.1"
+}
--- a/spec/fixtures/npm-license-expression/package.json
+++ b/spec/fixtures/npm-license-expression/package.json
@@ -1,6 +1,0 @@
-{
-  "name": "fake-package"
-  "description": "Fake package available under various licenses",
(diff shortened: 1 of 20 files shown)


spec/integration_spec.rb
```diff
@@ -177,6 +177,15 @@
...
         end
+
+        context 'with package.json license expression' do
+          let(:fixture) { 'npm-license-expression' }
+          let(:arguments) { { detect_packages: true } }
+
+          it 'matches other' do
+            expect(subject.license).to eql(other_license)
+          end
+        end
```
mlinksva: Maybe this test should go in `npm_bower_matcher_spec.rb` instead, skipping the fixture? Is suspect that's the case, but I'll leave here for now pending review.

benbalter: 👍

spec/fixtures/npm-license-expression/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "fake-package"
+  "description": "Fake package available under various licenses",
+  "license": "(MIT OR Apache-2.0 OR AGPL-3.0+)",
```
benbalter: It will take some reworking of the logic in the Project class, but in a subsequent iteration, it'd be cool to be able to parse the expression and return multiple licenses, where we can match the SPDX ID.

PR #241, opened 2017-11-23
Use LICENSE.spdx file if available

### LINKED ISSUE #85: Use LICENSE.spdx if available
I propose that if there's a file named "LICENSE.spdx" at the top level, this library should use the value of "PackageLicenseDeclared:" (if present) as the license of the package.

This lets people express _exactly_ what the license is using SPDX - so it can handle "or later", multiple licenses, or complex licensing situations.  Since it looks at "PackageLicenseDeclared", it _is_ from the developer, and since SPDX defines what the phrases mean exactly, I think you can assign a lot of confidence to this value.

This would give the licensee library a simple way to handle multiple licenses (per issue #57) with complex situations - just let the developer express it!

Don't get me wrong, I think that people should still include license text in the version control system.  But often that isn't enough.  Detecting the GPL version 2.0 license doesn't tell me if it's "GPL-2.0" (only) or "GPL-2.0+", and once multiple licenses are present, it's tricky to be confident that an automated tool can figure it out.  I think it's _safest_ to depend on a formal spec from the developer, and then use the other mechanisms when that fails.

I realize Ben already knows SPDX, but for readers of this issue who aren't aware of it, here is my [SPDX tutorial](https://github.com/david-a-wheeler/spdx-tutorial).

mlinksva:
Fixes #85 

Note because this is implemented as a package manager matcher, it won't be enabled unless package manager matching is turned on, which is for the command line by default, but not otherwise.

Here's output from a repo with a `LICENSE.spdx` file:

```
$ script/git-repo https://github.com/symphonyoss/clj-symphony
License: Apache License 2.0
Matched files: ["LICENSE", "README.md", "LICENSE.spdx"]
LICENSE:
  Confidence: 100.00%
  License: Apache License 2.0
README.md:
  Content hash: 4c9978cc3395868539a43bfe7037a750be9442f7
  Attribution: Copyright © 2016, 2017 Symphony Software Foundation
  Confidence: 90.00%
  License: Apache License 2.0
LICENSE.spdx:
  Confidence: 90.00%
  Matcher: Licensee::Matchers::Spdx
  License: Apache License 2.0
```

Or @david-a-wheeler's tutorial repo:

```
$ script/git-repo https://github.com/david-a-wheeler/spdx-tutorial
License: Other
Matched files: ["LICENSE.spdx"]
LICENSE.spdx:
  Confidence: 90.00%
  Matcher: Licensee::Matchers::Spdx
  License: Other
```
Other because licensee doesn't know about `CC-BY-3.0` (only 4.0), nor about `+`.

### DIFF
--- a/lib/licensee/matchers/spdx.rb
+++ b/lib/licensee/matchers/spdx.rb
@@ -1,0 +1,16 @@
+module Licensee
+  module Matchers
+    class Spdx < Licensee::Matchers::Package
+      # While we could parse the LICENSE.spdx file, prefer
+      # a lenient regex for speed and security. Moar parsing moar problems.
+      LICENSE_REGEX = /PackageLicenseDeclared:\s*([a-z\-0-9\. +()]+)\s*/i
+
+      private
+
+      def license_property
+        match = @file.content.match LICENSE_REGEX
+        match[1].downcase if match && match[1]
+      end
+    end
+  end
+end
--- a/lib/licensee/project_files/package_manager_file.rb
+++ b/lib/licensee/project_files/package_manager_file.rb
@@ -15,18 +15,21 @@ module Licensee
         'LICENSE.spdx' => [Matchers::Spdx]
       }.freeze
 
+      FILENAMES_SCORES = {
+        'package.json' => 1.0,
+        'LICENSE.spdx' => 1.0,
+        'DESCRIPTION'  => 0.9,
+        'dist.ini'     => 0.8,
+        'bower.json'   => 0.75
+      }.freeze
+
       def possible_matchers
         MATCHERS_EXTENSIONS[extension] || FILENAMES_EXTENSIONS[filename] || []
       end
 
       def self.name_score(filename)
-        return 1.0  if ['.gemspec', '.cabal'].include?(File.extname(filename))
-        return 1.0  if filename == 'package.json'
-        return 1.0  if filename == 'LICENSE.spdx'
-        return 0.8  if filename == 'dist.ini'
-        return 0.9  if filename == 'DESCRIPTION'
-        return 0.75 if filename == 'bower.json'
-        0.0
+        return 1.0 if ['.gemspec', '.cabal'].include?(File.extname(filename))
+        FILENAMES_SCORES[filename] || 0.0
       end
 
       private
--- a/lib/licensee/project_files/package_manager_file.rb
+++ b/lib/licensee/project_files/package_manager_file.rb
@@ -10,8 +10,9 @@ module Licensee
 
       # Hash of Filename => [possible matchers]
       FILENAMES_EXTENSIONS = {
-        'DESCRIPTION' => [Matchers::Cran],
-        'dist.ini'    => [Matchers::DistZilla]
+        'DESCRIPTION'  => [Matchers::Cran],
+        'dist.ini'     => [Matchers::DistZilla],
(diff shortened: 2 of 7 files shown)


lib/licensee/project_files/package_manager_file.rb
```diff
@@ -10,21 +10,26 @@ class PackageManagerFile < Licensee::ProjectFiles::ProjectFile
...
 
       def self.name_score(filename)
-        return 1.0  if ['.gemspec', '.cabal'].include?(File.extname(filename))
-        return 1.0  if filename == 'package.json'
-        return 0.8  if filename == 'dist.ini'
-        return 0.9  if filename == 'DESCRIPTION'
-        return 0.75 if filename == 'bower.json'
-        0.0
+        return 1.0 if ['.gemspec', '.cabal'].include?(File.extname(filename))
+        FILENAMES_SCORES[filename] || 0.0
```
benbalter: Nice.

lib/licensee/matchers/spdx.rb
```diff
@@ -0,0 +1,16 @@
+module Licensee
+  module Matchers
+    class Spdx < Licensee::Matchers::Package
+      # While we could parse the LICENSE.spdx file, prefer
+      # a lenient regex for speed and security. Moar parsing moar problems.
+      LICENSE_REGEX = /PackageLicenseDeclared:\s*([a-z\-0-9\. +()]+)\s*/i
```
wking: This will work in single-line cases, which are the most likely.  But the current tag/value format used in `.spdx` is not very clear about values wrapping over multiple lines (spdx/spdx-spec#44).  You may want a balanced-paren parser (reading to the end of the line or until any parens opened in that line are closed).  Or you can punt until the SPDX spec expresses a clearer opinion on this.

lib/licensee/project_files/package_manager_file.rb
```diff
@@ -10,21 +10,26 @@ class PackageManagerFile < Licensee::ProjectFiles::ProjectFile
...
-        'DESCRIPTION' => [Matchers::Cran],
-        'dist.ini'    => [Matchers::DistZilla]
+        'DESCRIPTION'  => [Matchers::Cran],
+        'dist.ini'     => [Matchers::DistZilla],
+        'LICENSE.spdx' => [Matchers::Spdx]
+      }.freeze
+
+      FILENAMES_SCORES = {
+        'package.json' => 1.0,
+        'LICENSE.spdx' => 1.0,
```
mlinksva: @wking this and `name_score` below are only used to prioritize which files to look at, see https://github.com/benbalter/licensee/blob/389c22e7e16378ed57c47b7b89e22e95281f8dab/lib/licensee/projects/project.rb#L84 they are not the confidence assigned to a license statement found in a package manager file. That's at https://github.com/benbalter/licensee/blob/389c22e7e16378ed57c47b7b89e22e95281f8dab/lib/licensee/matchers/package.rb#L14 (which specifies 90)

lib/licensee/matchers/spdx.rb
mlinksva: Thanks for mentioning this. I think it makes sense to punt until there's a clear opinion. I subscribed to the issue.

mlinksva:
Going ahead and merging, it's on par with other package matchers. Can improve based on clarifications from SPDX, future ability to parse expressions, etc, in future PRs.

Issue #207, opened 2017-06-26
FCPL detected as MPL-2.0


mlinksva:
https://github.com/benbalter/licensee/pull/205#issue-238585162

mlinksva:
Fixed by #212

Issue #225, opened 2017-09-04
NCSA License fails to be detected when fields are used correctly


narendasan:
I am trying to detect the NCSA License with ```licensee``` and I end up not getting any detection after I fill out all the fields 

Example:
Does not detect this License correctly 

```
University of Illinois/NCSA Open Source License 

Copyright (c) 2017 My Org. All rights reserved. 

Developed by: My Team 
              My Project  
              https://github.com/my-org/my-project
                  
Permission is hereby granted, free of charge, to any person 
obtaining a copy of this software and associated documentation files 
(the "Software"), to deal with the Software without restriction, 
including without limitation the rights to use, copy, modify, merge,
publish, distribute, sublicense, and/or sell copies of the Software, 
and to permit persons to whom the Software is furnished to do so, 
subject to the following conditions:

* Redistributions of source code must retain the above copyright notice, 
  this list of conditions and the following disclaimers.

* Redistributions in binary form must reproduce the above copyright 
  notice, this list of conditions and the following disclaimers in the 
  documentation and/or other materials provided with the distribution.

* Neither the names of My Team, My Project nor the names of its 
  contributors may be used to endorse or promote products derived from
  this Software without specific prior written permission.
  
THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS 
OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, 
FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE 
CONTRIBUTORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER 
LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, 
OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS WITH
THE SOFTWARE.
```

```licensee``` executable output:

```
License: Other
Matched files: ["LICENSE"]
LICENSE:
  Content hash: c18bbf3f6619c4d793bd4ea19bcaa854bb2cec13
  Attribution: Copyright (c) 2017 My Org. All rights reserved.
  License: Other
  Closest licenses:
    * NCSA similarity: 96.64%
```

The template I am using is the one provided on https://choosealicense.com/licenses/ncsa/

[...]

PR #242, opened 2017-11-24
scale dissimilarity with content length

### LINKED ISSUE #226: confidence_threshold not accurately reflected in dice matcher
Test case:

```ruby
#!/usr/bin/env ruby
require 'licensee'
Licensee.confidence_threshold = 85
p = Licensee.project(ARGV[0])
puts "#{p.license.key} #{p.matched_file.confidence}"
```

Running this against `https://raw.githubusercontent.com/pgbouncer/pgbouncer/master/COPYRIGHT`, for example, yields `other`. However, lowering the confidence threshold to 80 nets `isc 90.41095890410958`. This feels strange to me; a threshold of 90 should have even worked there, right? The highest integer threshold I'm able to use is 81.

This is not unique to that license file; it's just a good demonstration. I also saw this happening on another license that _should_ have matched at 85% confidence, yet it refused to identify until confidence was lowered around 60 or so.

mlinksva:
This is the core change:

```diff
     def similarity(other)
       overlap = (wordset & other.wordset).size
       total = wordset.size + other.wordset.size
-      100.0 * (overlap * 2.0 / total)
+      100.0 * ((overlap * 2.0 / total)**Math.log10(wordset.size))
     end
```
Short content or highly similar content will have barely changed similarity, but the longer the content, the more small increments in dissimilarity is punished.

This allows us to not rely on max_delta to filter out long texts with relatively high pre-scaled similarity. Max delta can be infinite and all tests still pass, no false positives on significant but small relative to size changes in long licenses (eg #192 and #207 which are fixed, but without this scaling do rely on max_delta filtering them out as candidates).

Also test that content with given similarity will be included in similar licenses, given confidence threshold (above 50) equal to given similarity for more intuitive results; fixes #226. Note this test required a few other changes because licensee previously couldn't respond to dynamic changes in confidence threshold.

### DIFF
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -28,7 +28,7 @@ module Licensee
     # Number of characters that could be added/removed to still be
     # considered a potential match
     def max_delta
-      @max_delta ||= (length * Licensee.inverse_confidence_threshold).to_i
+      @max_delta ||= calculate_max_delta
     end
 
     # Given another license or project file, calculates the difference in length
@@ -41,7 +41,7 @@ module Licensee
     def similarity(other)
       overlap = (wordset & other.wordset).size
       total = wordset.size + other.wordset.size
-      100.0 * (overlap * 2.0 / total)
+      100.0 * ((overlap * 2.0 / total)**Math.log10(total))
     end
 
     # SHA1 of the normalized content
@@ -124,6 +124,10 @@ module Licensee
     end
 
     private
+
+    def calculate_max_delta
+      (length * Licensee.inverse_confidence_threshold).to_i * 2
+    end
 
     def strip_title(string)
       strip(string, ContentHelper.title_regex)
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -16,7 +16,7 @@ module Licensee
     # A set of each word in the license, without duplicates
     def wordset
       @wordset ||= if content_normalized
-        content_normalized.scan(/[\w']+/).to_set
+        content_normalized.scan(/(?:\w(?:'s|(?<=s)')?)+/).to_set
       end
     end
 
@@ -86,6 +86,7 @@ module Licensee
         string = strip_all_rights_reserved(string)
         string, _partition, _instructions = string.partition(END_OF_TERMS_REGEX)
         string = strip_markup(string)
+        string = normalize_quotes(string)
         strip_whitespace(string)
       end
 
@@ -170,6 +171,13 @@ module Licensee
 
     def strip(string, regex)
       string.gsub(regex, ' ').squeeze(' ').strip
+    end
+
+    # Replace all single quotes with double quotes
+    # Single versus double quotes don't alter the meaning, and it's easier to
+    # strip double quotes if we still want to allow possessives
+    def normalize_quotes(string)
+      string.gsub(/\s'([\w -]*?\w)'/, ' "\1"')
     end
   end
 end
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -28,7 +28,7 @@ module Licensee
     # Number of characters that could be added/removed to still be
     # considered a potential match
     def max_delta
-      @max_delta ||= calculate_max_delta
+      (length * Licensee.inverse_confidence_threshold).to_i * 2
     end
 
     # Given another license or project file, calculates the difference in length
@@ -37,11 +37,16 @@ module Licensee
     end
 
     # Given another license or project file, calculates the similarity
-    # as a percentage of words in common
+    # as a percentage of words in common, minus a penalty that increases
+    # as percentage of words in common decreases and wordset size increases
+    # so that false positives for long licnses are ruled out by this score
+    # alone. The penality is incurred by taking the decimal proportion to a
+    # power (**) which is itself scaled by wordset size (log10 of 100 is 2,
+    # 1000 is 3).
     def similarity(other)
       overlap = (wordset & other.wordset).size
       total = wordset.size + other.wordset.size
-      100.0 * ((overlap * 2.0 / total)**Math.log10(total))
+      100.0 * ((overlap * 2.0 / total)**Math.log10(wordset.size))
     end
 
     # SHA1 of the normalized content
@@ -124,10 +129,6 @@ module Licensee
     end
 
     private
-
-    def calculate_max_delta
-      (length * Licensee.inverse_confidence_threshold).to_i * 2
-    end
 
     def strip_title(string)
       strip(string, ContentHelper.title_regex)
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -4,13 +4,14 @@ require 'digest'
 module Licensee
   module ContentHelper
     DIGEST = Digest::SHA1
-    END_OF_TERMS_REGEX = /^\s*end of terms and conditions\s*$/i
+    END_OF_TERMS_REGEX = /^[\s#*_]*end of terms and conditions\s*$/i
     HR_REGEX = /[=\-\*][=\-\*\s]{3,}/
     ALT_TITLE_REGEX = License::ALT_TITLE_REGEX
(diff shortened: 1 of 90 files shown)


lib/licensee/content_helper.rb
```diff
@@ -28,7 +28,7 @@ def length
     # Number of characters that could be added/removed to still be
     # considered a potential match
     def max_delta
-      @max_delta ||= (length * Licensee.inverse_confidence_threshold).to_i
+      @max_delta ||= calculate_max_delta
```
benbalter: Curious why you broke this out into a new method? If rubocop was complaining about line length, you can wrap it in a `begin`/`end` block and put it on its own line.

lib/licensee/content_helper.rb
```diff
@@ -41,7 +41,7 @@ def length_delta(other)
     def similarity(other)
       overlap = (wordset & other.wordset).size
       total = wordset.size + other.wordset.size
-      100.0 * (overlap * 2.0 / total)
+      100.0 * ((overlap * 2.0 / total)**Math.log10(total))
```
benbalter: What does `**` do?

spec/licensee/matchers/dice_matcher_spec.rb
```diff
@@ -22,12 +22,12 @@
 
   it 'sorts licenses by similarity' do
     expect(subject.licenses_by_similiarity[0]).to eql([gpl, 100.0])
-    expect(subject.licenses_by_similiarity[1]).to eql([agpl, 95.73361082206036])
+    expect(subject.licenses_by_similiarity[1]).to eql([agpl, 86.66032027067445])
```
benbalter: Do we know why this is changing?

spec/licensee/matchers/dice_matcher_spec.rb
```diff
@@ -80,4 +80,39 @@
...
+      context "the #{license.key} license" do
+        let(:content) { license.content }
+
+        nearest =
+          Licensee
+          .licenses(hidden: true)
+          .reject { |other| other.pseudo_license? || other.key == license.key }
+          .collect { |other| [other, license.similarity(other)] }
+          .max_by { |other_similarity| other_similarity[1] }
+        next if nearest[1] < 50
```
benbalter: What's going on here? I think this logic could use a comment.

spec/licensee/matchers/dice_matcher_spec.rb
```diff
@@ -80,4 +80,39 @@
...
+
+        it "matches #{nearest[0].key}" do
+          Licensee.confidence_threshold = nearest[1].floor
+
+          matcher = Licensee::Matchers::Dice.new(file)
+          similars = matcher.licenses_by_similiarity.map { |s| s[0] }
+
+          expect(similars).to include nearest[0]
```
benbalter: What are you trying to test here? That potential licenses include licenses above the set confidence threshold?

spec/licensee/content_helper_spec.rb
```diff
@@ -39,7 +39,7 @@ def initialize(content = nil)
     let(:content) { 'license' * 1000 }
 
     it 'returns the max delta' do
-      expect(subject.max_delta).to eql(140)
+      expect(subject.max_delta).to eql(280)
```
benbalter: Should we add a test to ensure a short license has the expected max delta?

benbalter:
Excited to see this problem finally having a solid solution!

lib/licensee/content_helper.rb
mlinksva: To make it easier to monkeypatch to get the uncached version in a test that varies confidence level (which max_delta needs to respond to). I'll just remove the cache; this is not a heavy calculation.

lib/licensee/content_helper.rb
mlinksva: Power operator. Base similarity to power of (log10 of wordset size). That's how dissimilarity is  scaled in this proposal.

I'm going to change total to wordset.size as the argument to log10 though, since there's no need for the scaling to be sensitive to the size of the candidate text, and it makes the scaling slightly less aggressive.

spec/licensee/matchers/dice_matcher_spec.rb
mlinksva: Yes, GPL and AGPL are long licenses, so they are getting hit hard by dissimilarity scaling with size mentioned at https://github.com/benbalter/licensee/pull/242#discussion_r153684000

spec/licensee/matchers/dice_matcher_spec.rb
mlinksva: Yes.

spec/licensee/content_helper_spec.rb
mlinksva: Added.

spec/licensee/matchers/dice_matcher_spec.rb
mlinksva: Adding a comment here (and a few others), but in short we're finding the most similar license to the one we're looking at (we're going to see if it is in the set of similar licenses, not cut off from comparison due to max delta), but not if confidence is set to less than 50 in which case all bets are off, and who would want to set confidence so low anyway (in fact only one pair fails without this).

mlinksva:
Here's the impact of this proposal on hypothetical license texts that are 99% similar to each of the licenses cataloged -- the last column is with dissimilarity scaled -- increasing with size.

```
License            Length   MaxDelta Wordset  Similarity Scaled  
wtfpl                   483        6       37    99.00    98.44
isc                     744       13       66    99.00    98.19
zlib                    856       16       77    99.00    98.12
postgresql              940       17       86    99.00    98.07
mit                    1069       20       91    99.00    98.05
unlicense              1210       24      111    99.00    97.97
bsd-2-clause           1317       24      100    99.00    98.01
bsl-1.0                1338       25      111    99.00    97.97
bsd-3-clause           1512       28      116    99.00    97.95
bsd-3-clause-clear     1655       31      125    99.00    97.91
ncsa                   1663       29      117    99.00    97.94
ms-pl                  2619       51      150    99.00    97.84
ms-rl                  3005       59      168    99.00    97.79
ofl-1.1                4345       83      253    99.00    97.61
cc0-1.0                6555      129      359    99.00    97.46
lgpl-3.0               7651      143      300    99.00    97.54
artistic-2.0           8894      170      370    99.00    97.45
afl-3.0               10294      204      460    99.00    97.36
osl-3.0               10296      204      455    99.00    97.36
ecl-2.0               11066      196      435    99.00    97.38
apache-2.0            11357      181      412    99.00    97.41
epl-1.0               11514      223      457    99.00    97.36
eupl-1.1              13121      260      532    99.00    97.30
mpl-2.0               16725      294      532    99.00    97.30
gpl-2.0               18092      293      602    99.00    97.25
cc-by-4.0             18646      332      566    99.00    97.27
lppl-1.3c             19039      358      656    99.00    97.21
cc-by-sa-4.0          20127      358      579    99.00    97.26
lgpl-2.1              26526      474      792    99.00    97.13
agpl-3.0              34520      629      951    99.00    97.05
gpl-3.0               35147      631      971    99.00    97.04
```

You'll note that effectively the similarity threshold is increased to approaching 99% for short licenses, even higher for long licenses. That's not necessarily a bad thing, but if we go with this proposal, it might make sense to change the threshold from 98 to 97, because effectively we'd be grading on a scale.

I'm also struck that we probably could have a stronger in-the-wild test dataset which could inform the default threshold. One idea for getting that is to leverage the 0-3 example deployments of each cataloged licenses available as license metadata. I [recently added a test over at choosealicense.com](https://github.com/github/choosealicense.com/pull/555) to ensure that those examples are detected by the current version of licensee, so this would not be difficult to do.

mlinksva:
Had an idea to address #225 by removing mostly* fill-in fields from the similarly calculation, implemented in de65a8b

I started by implementing separate from this branch, but ran into max_delta prematurely filtering out candidates. I'm not sure how big of a barrier that is though, might be worth implementing separately.

* mostly because the filled in license could have for example `Ben Balter` (2 words) for `fullname` (1 word), which leaves an unaccounted for non-overlapping word in text we're looking for a match against.

PR #246, opened 2017-12-13
Extend gemspec pattern matching

jonabc:
Found a few issues with parsing the `license` field out of a gemspec.

1. The gem specification allows for a `licenses` array field that wasn't found and captured.
2. The pattern didn't match on frozen strings, e.g. `"mit".freeze`

Ruby allows regex's to be interpolated inside other regex's which makes the array regex much easier to understand and work with.  It also made it simpler to ensure values were treated equally among all of the regular expressions.

The only captured group in these regex's is the license text.  All other groups are non-capturing with `(?: ...)`.  Meaning in an array like `["mit", "bsd-3-clause"]`, `match[1] == "mit"` and `match[2] == "bsd-3-clause"`

Only the first array value is used, to match current functionality.

### DIFF
--- a/lib/licensee/matchers/gemspec.rb
+++ b/lib/licensee/matchers/gemspec.rb
@@ -28,10 +28,18 @@ module Licensee
         match = @file.content.match LICENSE_REGEX
         return match[1].downcase if match && match[1]
 
-        # this will capture each value in the license array
-        # for now only return the first match
+        # check for a licenses array property
+        licenses = license_array_property
+        return unless licenses
+
+        # use 'other' if array contains multiple licenses
+        return 'other' unless licenses.size == 1
+        licenses[0]
+      end
+
+      def license_array_property
         match = @file.content.match LICENSE_ARRAY_REGEX
-        match[1].downcase if match && match[1]
+        match.captures.compact.map(&:downcase) if match
       end
 
       def declarations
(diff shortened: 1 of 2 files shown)


mlinksva:
If there are multiple values in a licenses array I suspect the best match for current functionality would be to return `other`. That's what licensee does when it encounters multiple licenses, whether from different (eg README and license file) or the same matchers, eg when it finds a license expression in npm packages (though it doesn't attempt to parse out the multiple licenses).

lib/licensee/matchers/gemspec.rb
```diff
@@ -1,18 +1,36 @@
...
 
       private
 
       def license_property
         match = @file.content.match LICENSE_REGEX
+        return match[1].downcase if match && match[1]
+
+        # this will capture each value in the license array
+        # for now only return the first match
+        match = @file.content.match LICENSE_ARRAY_REGEX
```
benbalter: Ideally, we'd abstract things such that a project file could have multiple licenses, but for now, I agree with @mlinksva that "other" would be the consistent response here (but I'd leave the other logic intact to allow for that). Perhaps expose a `matches` method for now (in addition to `match`)?

jonabc:
@benbalter @mlinksva thanks for taking a look!  I've updated per feedback, though I'm not sure why the `dependency-ci` job failed 🙁

lib/licensee/matchers/gemspec.rb
jonabc: I avoided adding any new fields at a broad scope because I wasn't sure how they would be used.  It definitely seems worth thinking about for the future.

benbalter:
>  I'm not sure why the dependency-ci job failed

Probably just added that to test the service. No idea. Safe to ignore.

PR #249, opened 2017-12-20
Strip markdown prior to comparison

benbalter:
This PR updates `ContentHelper#content_normalized` to strip any character that's not a letter, a number, a period or a hyphen, meaning markup isn't included for length-comparison purposes.

We could, alternatively, just strip the markup before we calculate the length, but I thought that this better normalized the content (E.g. for the exact matcher). The markup is stripped prior to wordset creation, meaning it's not used in the Dice matcher either way.

My thinking was that words, numbers, period, and dashes were legally significant, but other markup (e.g. formatting) was not.

Thoughts? If we want to go this route, I can update test expectations.

Fixes https://github.com/benbalter/licensee/issues/247.

### DIFF
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -4,13 +4,14 @@ require 'digest'
 module Licensee
   module ContentHelper
     DIGEST = Digest::SHA1
-    END_OF_TERMS_REGEX = /^\s*end of terms and conditions\s*$/i
+    END_OF_TERMS_REGEX = /^[\s#]*end of terms and conditions\s*$/i
     HR_REGEX = /[=\-\*][=\-\*\s]{3,}/
     ALT_TITLE_REGEX = License::ALT_TITLE_REGEX
     ALL_RIGHTS_RESERVED_REGEX = /\Aall rights reserved\.?$/i
     WHITESPACE_REGEX = /\s+/
     MARKDOWN_HEADING_REGEX = /\A\s*#+/
     VERSION_REGEX = /\Aversion.*$/i
+    MARKUP_REGEX = /[^\w'\.\-]+/
 
     # A set of each word in the license, without duplicates
     def wordset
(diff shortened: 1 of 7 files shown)


lib/licensee/content_helper.rb
```diff
@@ -4,13 +4,14 @@
 module Licensee
   module ContentHelper
     DIGEST = Digest::SHA1
-    END_OF_TERMS_REGEX = /^\s*end of terms and conditions\s*$/i
+    END_OF_TERMS_REGEX = /^[\s#]*end of terms and conditions\s*$/i
```
mlinksva: I wonder if it's worth catching markdown bold/italics here as well as header.

lib/licensee/content_helper.rb
```diff
@@ -4,13 +4,14 @@
...
     HR_REGEX = /[=\-\*][=\-\*\s]{3,}/
     ALT_TITLE_REGEX = License::ALT_TITLE_REGEX
     ALL_RIGHTS_RESERVED_REGEX = /\Aall rights reserved\.?$/i
     WHITESPACE_REGEX = /\s+/
     MARKDOWN_HEADING_REGEX = /\A\s*#+/
     VERSION_REGEX = /\Aversion.*$/i
+    MARKUP_REGEX = /[^\w'\.\-,]+/
```
mlinksva: Is the `'` there because it can make a word possessive, but not `"` because it can't change meaning?

If a comma is considered meaningful, shouldn't a semicolon also be? What about parens?

benbalter: You're absolutely right (and that was my original thinking). I checked the SPDX license matching spec, and it says formatting should be ignored, but punctuation should not. Updated with a blacklist of Markdown markup characters via 08f3779.

lib/licensee/content_helper.rb
benbalter: Good call. 7bd1aa6.

PR #251, opened 2017-12-24
Add support for scanning GitHub projects

nickcampbell18:
I had a requirement to scan remote Github projects to try and find their license.

So I took inspiration from the filesystem implementation (`Licensee::Projects::FSProject`) and built one which uses the GitHub API to resolve each of the files just in the root directory.

Is this something you might be interesting in having in the main library? If so, I can add some tests, error handling and probably a better HTTP adapter that can make requests in parallel. It would definitely be worth considering what the public interface would look like (I'm currently just using this gem from within another Ruby application, not sure if/how the CLI would change).

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -2,18 +2,22 @@
 
 ## Command line usage
 
-1. `cd` into a project directory
-2. Execute the `licensee` command
+This gem includes an executable which can be run using the `licensee [PATH]` command,
+where `[PATH]` is:
 
-You'll get an output that looks like:
+* A directory, for example: `licensee vendor/gems/activesupport`
+* A file, for example: `licensee LICENSE.txt`
+* A GitHub repository, for example: `licensee https://github.com/facebook/react`
+
+If you don't specify any arguments, `licensee` will just scan the current directory.
+
+In all cases, you'll get an output that looks like:
 
 ```
 License: MIT
 Confidence: 98.42%
 Matcher: Licensee::GitMatcher
 ```
-
-Alternately, `licensee <directory>` will treat the argument as the project directory, and `licensee <file>` will attempt to match the individual file specified, both with output that looks like the above.
 
 ## License Ruby API
 
@@ -35,6 +39,24 @@ license.meta["description"]
 
 license.meta["permissions"]
 => ["commercial-use","modifications","distribution","private-use"]
+```
+
+If you wish to scan private GitHub repositories, or are hitting API rate limits, you can configure the embedded [Octokit](https://github.com/octokit/octokit.rb)
+client using environment variables, for example:
+
+```sh
+OCTOKIT_ACCESS_TOKEN=abc123 licensee https://github.com/benbalter/licensee
+```
+
+Octokit can also be configured using standard module-level configuration:
+
+```ruby
+# see https://github.com/octokit/octokit.rb#configuring-module-defaults
+Octokit.configure do |c|
+  c.access_token = "<your 40 char token>"
+end
+
+license = Licensee.license "https://github.com/benbalter/licensee"
 ```
 
 ## Advanced API usage
--- a/lib/licensee/projects.rb
+++ b/lib/licensee/projects.rb
@@ -3,5 +3,6 @@ module Licensee
     autoload :Project, 'licensee/projects/project'
     autoload :FSProject, 'licensee/projects/fs_project'
     autoload :GitProject, 'licensee/projects/git_project'
+    autoload :GitHubProject, 'licensee/projects/github_project'
   end
 end
--- a/lib/licensee/projects/github_project.rb
+++ b/lib/licensee/projects/github_project.rb
@@ -1,0 +1,46 @@
+# GitHub project
+#
+# Analyses a remote GitHub repository for license information
+#
+# Only the root directory of a repository will be scanned because every
+# `#load_file(..)` call incurs a separate API request.
+
+require 'octokit'
+
+module Licensee
+  module Projects
+    class GitHubProject < Licensee::Projects::Project
+      # If there's any trailing data (e.g. `.git`) this pattern will ignore it:
+      # we're going to use the API rather than clone the repo.
+      GITHUB_REPO_PATTERN = %r{https://github.com/([^\/]+\/[^\/\.]+).*}
+
+      class RepoNotFound < StandardError; end
+
+      def initialize(github_url, **args)
+        @repo = github_url[GITHUB_REPO_PATTERN, 1]
+        raise ArgumentError, "Not a github URL: #{github_url}" unless @repo
+        super(**args)
+      end
+
+      attr_reader :repo
+
+      private
+
+      def files
+        @files ||= contents.map { |data| { name: data[:name], dir: '/' } }
+      rescue Octokit::NotFound
+        raise RepoNotFound,
+              "Could not load GitHub repo #{repo}, it may be private or deleted"
+      end
+
+      def load_file(file)
+        Octokit.contents(@repo, path:   file[:name],
+                                accept: 'application/vnd.github.v3.raw')
+      end
+
+      def contents
+        Octokit.contents(@repo).select { |data| data[:type] == 'file' }
+      end
+    end
+  end
+end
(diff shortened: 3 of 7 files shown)


benbalter:
@nickcampbell18 Nice! I like it. FYI, we also have `script/git-repo` which performs a shallow clone of the Git repository, but think this could be more efficient for debugging.

If you're interested in getting this PR merged, I'd suggest taking a look at Octokit, which should have helper functions for all the endpoints we'd need.

If I'm reading correctly, it looks like we'll want to make one call to get the list of root files, and then another for each file's contents that we need, which shouldn't be too bad (and we can include instructions for passing an API key as well).

nickcampbell18:
Thanks! I've made the changes and updated the usage instructions with an example of how you might want to run this, does this make sense to you? If so I'll push on and write the tests & add some better error handling.

> I'd suggest taking a look at Octokit

I wasn't sure about bringing in a dependency for just this one API call, but it does make the code a bit more readable!

docs/usage.md
```diff
@@ -2,19 +2,23 @@
...
 
-1. `cd` into a project directory
-2. Execute the `licensee` command
+This gem includes an executable which can be run using the `licensee [PATH]` command,
+where `[PATH]` is:
 
-You'll get an output that looks like:
+* A directory, for example: `licensee vendor/gems/activesupport`
+* A file, for example: `licensee LICENSE.txt`
+* A Github repository, for example: `licensee https://github.com/facebook/react`
```
benbalter: Git**H**ub :smile:

docs/usage.md
```diff
@@ -37,6 +41,18 @@ license.meta["permissions"]
...
+
+```ruby
+# see https://github.com/octokit/octokit.rb#configuring-module-defaults
+Octokit.configure do |c|
+  c.access_token = "<your 40 char token>"
+end
+
+license = Licensee.license "https://github.com/benbalter/licensee"
+```
+
```
benbalter: It may also be worth noting, that for command-line invocations, you can pass it as an environmental variable, e.g.:

```
OCTOKIT_ACCESS_TOKEN=abc123 licensee foo/bar
```

docs/usage.md
```diff
@@ -37,6 +41,18 @@ license.meta["permissions"]
 => ["commercial-use","modifications","distribution","private-use"]
 ```
 
+If you wish to scan private Github repositories, you can configure the embedded [Octokit](https://github.com/octokit/octokit.rb) client
```
benbalter: > If you wish to scan private Github repositories

I'd add something along the lines of "or are hitting API rate limits"

lib/licensee/projects.rb
```diff
@@ -3,5 +3,6 @@ module Projects
     autoload :Project, 'licensee/projects/project'
     autoload :FSProject, 'licensee/projects/fs_project'
     autoload :GitProject, 'licensee/projects/git_project'
+    autoload :GithubProject, 'licensee/projects/github_project'
```
benbalter: Given that FSProject is FSProject and not FsProject, I'm wondering if this should be GitHubProject for consistency?

lib/licensee/projects/github_project.rb
```diff
@@ -0,0 +1,43 @@
...
+
+module Licensee
+  module Projects
+    class GithubProject < Licensee::Projects::Project
+      # If there's any trailing data (e.g. `.git`) this pattern will ignore it:
+      # we're going to use the API rather than clone the repo.
+      GITHUB_REPO_PATTERN = %r{https://github.com/([^\/]+\/[^\/\.]+).*}
+
+      def initialize(github_url, **args)
+        @repo_name = github_url[GITHUB_REPO_PATTERN, 1]
```
benbalter: Had no idea you could do this. Nice.

lib/licensee/projects/github_project.rb
```diff
@@ -0,0 +1,43 @@
...
+          {
+            name:         data[:name],
+            contents_url: data[:download_url],
+            dir:          '/'
+          }
+        end
+      end
+
+      def load_file(file)
+        Octokit.get URI(file[:contents_url])
```
benbalter: I believe we can use `Octokit.contents @repo_name, path: file[:name]` here to rely on Octokit internals a bit more (and save us from needing to store the `contents_url` in the hash).

lib/licensee/projects/github_project.rb
```diff
@@ -0,0 +1,43 @@
...
+      end
+
+      def fetch_files
+        Octokit.contents(@repo_name).select { |data| data[:type] == 'file' }
```
benbalter: Does `repo` make more sense than `repo_name`? Whatever we call it, I think we can use `attr_accessor` to mirror the other project classes's style.

lib/licensee/projects/github_project.rb
benbalter: Is there any reason we'd need to memoize/cache this method? I don't think so (and if so, could be a good future improvement).

nickcampbell18:
Thanks, I've made those changes and added tests, worth another look whenever you're ready!

spec/licensee/projects/github_project_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    context 'with a GitHub git URI' do
+      let(:github_url) { 'https://github.com/benbalter/licensee.git' }
+
+      it 'should set @repo, stripping the trailing extension' do
+        expect(instance.repo).to eq('benbalter/licensee')
+      end
+    end
+
+    context 'with a non-GitHub URI' do
+      let(:github_url) { 'https://gitlab.com/benbalter/licensee' }
```
benbalter: Just making a note, I know GitLab uses Licensee, if they/anyone would like more flexibility here, glad to add it, as long as we can do so safely. :smile:

spec/licensee/projects/github_project_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    context 'with a local folder' do
+      let(:github_url) { fixture_path('mit') }
+
+      it 'should raise an ArgumentError' do
+        expect { instance }.to raise_error(ArgumentError)
+      end
+    end
+  end
+
+  subject { described_class.new(github_url) }
```
benbalter: Very minor, but I believe you can put this in the top-level RSpec block, rather than in each context/describe.

spec/licensee/projects/github_project_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+
+  let(:repo) { 'benbalter/licensee' }
+  let(:github_url) { 'https://github.com/benbalter/licensee' }
```
benbalter: I don't know if/why we'd change this, but to make the test a bit less susceptible to external change (e.g., if we move the project to `licensee/licensee`, I'd suggest making the references to `benbalter/licensee` point to a single `#{repo}` (from the let statement), wherever possible (e.g., it won't be possible in the before block).

spec/licensee/projects/github_project_spec.rb
```diff
@@ -0,0 +1,140 @@
...
+    File.read(fixture_path('mit/README.md'))
+  end
+  let(:license_file) { File.read(fixture_path('mit/LICENSE.txt')) }
+
+  context 'when the repo exists' do
+    before do
+      allow(Octokit)
+        .to receive(:contents)
+        .with('benbalter/licensee')
+        .and_return([
```
benbalter: Would never have thought to do it this way. For other projects, I've used VCR, which stubs the HTTP API response, rather than the library method, in case the library changes its behavior. Out of scope for this PR, and glad to implement as a follow up.

spec/licensee/projects/github_project_spec.rb
```diff
@@ -0,0 +1,140 @@
+RSpec.describe Licensee::Projects::GitHubProject do
```
benbalter: One idea to make this a tighter integration tests with the other project types would be to move one of the fixtures to a submodule and point the test directly to that repo when exercising this project type (or locally for the other two).

benbalter:
A few very minor stylistic comments on the tests that were easy enough to implement. Really excited for this @nickcampbell18! Thank you!

Issue #172, opened 2017-03-04
CC0 not being recognized when following creativecommons.org instructions


C4K3:
First I'm sorry if this is the wrong place, the root of this issue is that choosealicense uses a slightly edited version of CC0 compared to the one CC distributes, so this might be the wrong place.

Currently licensee cannot detect CC0 added as per the instructions at https://wiki.creativecommons.org/wiki/CC0_FAQ#May_I_apply_CC0_to_computer_software.3F_If_so.2C_is_there_a_recommended_implementation.3F instead giving the following output
```
License file: COPYING
License hash: f5b7dd1f4796769f765ab4b8d40304c2bf1bcc47

Here's the closest licenses:
* LGPL-3.0 similarity: 29.60%
```
The reason for this appears to mostly be the disclaimer, if you remove that then it gets correctly recognized
```
License file: COPYING
License hash: f2d25c552d35d3e35b5041105b993da5a4b058c9
License: Creative Commons Zero v1.0 Universal
Confidence: 98.59%
Method: Licensee::Matchers::Dice
```

mlinksva:
The license text used by licensee was introduced in https://github.com/github/choosealicense.com/pull/126 (licensee vendors choosealicense.com license texts) presumably by copying and pasting from https://creativecommons.org/publicdomain/zero/1.0/legalcode though I'm not sure how to explain the link at the end. Probably should be updated with https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt or https://spdx.org/licenses/CC0-1.0.html ... latter marks disclaimer as optional and doesn't have first line "Creative Commons Legal Code" (probably should be present there but also marked as optional).

I'm not sure what the right solution for licensee is at this point given it doesn't know anything about optional sections, but I made https://github.com/github/choosealicense.com/pull/488 updating the license text from https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt in case that's useful for testing.

red-001:
why bot

anowlcalledjosh:
Any chance this could be reopened? The issue of CC0 not being recognised is still happening.

PR #253, opened 2017-12-28
Detect CC0-1.0 both as published by CC/SPDX and choosealicense.com 2013-2019

### LINKED ISSUE #243: Proper CC0 license is not recognized
This is a follow-up on #172 and github/choosealicense.com#488.

CC0 version distributed by Creative Commons is not recognized: https://creativecommons.org/publicdomain/zero/1.0/legalcode.txt

For example, see: https://github.com/yegortimoshenko/choosealicense.com-488/blob/8c9e0df02cb6fb74667298781c644116ac89b544/LICENSE

mlinksva:
Ensure that we recognize both text to be merged in https://github.com/github/choosealicense.com/pull/488 and text published at choosealicense.com and used by licensee since 2013.

Added tests, naive stripping of optional line at beginning (from CC source) and end (from choosealicense-2013).

Not sure how to deal with large block of optional text in source:

```
    CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE
    LEGAL SERVICES. DISTRIBUTION OF THIS DOCUMENT DOES NOT CREATE AN
    ATTORNEY-CLIENT RELATIONSHIP. CREATIVE COMMONS PROVIDES THIS
    INFORMATION ON AN "AS-IS" BASIS. CREATIVE COMMONS MAKES NO WARRANTIES
    REGARDING THE USE OF THIS DOCUMENT OR THE INFORMATION OR WORKS
    PROVIDED HEREUNDER, AND DISCLAIMS LIABILITY FOR DAMAGES RESULTING FROM
    THE USE OF THIS DOCUMENT OR THE INFORMATION OR WORKS PROVIDED
    HEREUNDER.
```

Stripping that entire string regardless of how it is wrapped would be one straightforward option.

Fixes  #243 which is a duplicate of #172


lib/licensee/content_helper.rb
```diff
@@ -12,6 +12,18 @@ module ContentHelper
     MARKDOWN_HEADING_REGEX = /\A\s*#+/
     VERSION_REGEX = /\Aversion.*$/i
     MARKUP_REGEX = /[#_*=~\[\]()`|>]+/
+    CC_LEGAL_CODE_REGEX = /^\s*Creative Commons Legal Code\s*$/i
+    CC0_INFO = 'For more information, please see\s*' +
+               '<http://creativecommons.org/publicdomain/zero/1.0/>\s*'.freeze
+    CC0_INFO_REGEX = /#{CC0_INFO}/im
+    CC_DISCLAIMER = %(CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES
```
wking: You may want to label this `CC0_DISCLAIMER` or some such.  The other CC licenses have similar disclaimers, but adds additional wording for “information”.  For example, compare the [CC-BY-3.0][1] disclaimer.  The [CC-BY-4.0][2] disclaimer is even more different, dropping caps, adding `("Creative Commons")`, etc.

[1]: https://creativecommons.org/licenses/by/3.0/legalcode.txt
[2]: https://creativecommons.org/licenses/by/4.0/legalcode.txt

mlinksva:
I'm baffled as to why a wholly unrelated test just started failing (it doesn't locally):

```
  1) command line invocation when given a repo URL detects the file's license
     Failure/Error: expect(stdout).to match('License: MIT License')
       expected "" to match "License: MIT License"
     # ./spec/bin_spec.rb:69:in `block (3 levels) in <top (required)>'
```

I'll try restarting the CI test jobs later in hopes its something transient in the CI environment.

benbalter:
> I'm baffled as to why a wholly unrelated test just started failing (it doesn't locally):

Ugh. I suspect it's hitting the API rate limit. I just added that test today, but it's largely an integration test. I think it's safe to remove.

mlinksva:
OK, it might be ugly, but it works. There are probably ways to make it slightly less ugly; review requested.

Longer term and more generally I'd like to see if extracting `<optional>` texts from https://github.com/spdx/license-list-XML rather than hardcoding them here might be workable. Beyond that, explore a new matcher based entirely on following the markup in that repo.

wking:
Does this need to land before github/choosealicense.com#488 is merged?

mlinksva:
More or less. I want to be sure that licensee will be able to recognize both CC0-1.0 texts (via this PR or another solution) before merging https://github.com/github/choosealicense.com/pull/488 so that the next time licensee vendors choosealicense.com suddenly CC0-1.0 detection fails for any of the [~30k repos with the text from choosealicense.com to date](https://github.com/spdx/license-list-XML/pull/489).

lib/licensee/content_helper.rb
```diff
@@ -158,6 +171,13 @@ def strip_all_rights_reserved(string)
       strip(string, ALL_RIGHTS_RESERVED_REGEX)
     end
 
+    def strip_cc0_optional(string)
+      return string unless string.include? 'cc0'
+      string = strip(string, CC_LEGAL_CODE_REGEX)
```
benbalter: If this is an alternate title, I believe if you add a reference in https://github.com/benbalter/licensee/blob/1553efdd953953818272a60b3ac2185c8aa0e4c4/lib/licensee/license.rb#L88-L92, it should be stripped by existing normalizers.

For context, here's the existing cc0 title regex:

```
/(?i-mx:creative\ commons\ zero\,?\ (?:version\ |v)?1\.0\ universal)|cc0\-1\.0/
```

lib/licensee/content_helper.rb
```diff
@@ -12,6 +12,18 @@ module ContentHelper
     MARKDOWN_HEADING_REGEX = /\A\s*#+/
     VERSION_REGEX = /\Aversion.*$/i
     MARKUP_REGEX = /[#_*=~\[\]()`|>]+/
+    CC_LEGAL_CODE_REGEX = /^\s*Creative Commons Legal Code\s*$/i
+    CC0_INFO = 'For more information, please see\s*' +
+               '<http://creativecommons.org/publicdomain/zero/1.0/>\s*'.freeze
+    CC0_INFO_REGEX = /#{CC0_INFO}/im
+    CC0_DISCLAIMER = %(CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES
```
benbalter: Can we dynamically pull this from the license text, e.g., via regex rather than hard coding it twice in the project?

lib/licensee/content_helper.rb
mlinksva: It's really not. Titles are used by the reference matcher (for README files) and `CREATIVE COMMONS LEGAL CODE` doesn't make any sense there. It's some cruft added to the license text by the license steward, for better or worse, that is optional/often isn't deployed in the wild, but often is.

lib/licensee/content_helper.rb
mlinksva: We could. Seems more complicated than just repeating, but could be used to handle other optional sections as well. Probably a good idea. I'll try adding in a bit.

mlinksva:
@benbalter just a ping to review this again when you get the chance. As mentioned at https://github.com/github/choosealicense.com/pull/488#issuecomment-379062833 SPDX 3.1 will account for the CC0-1.0 text variation introduced in choosealicense.com and licensee; it'd be nice to reciprocate by moving to the standard text in choosealicense.com and licensee. 😄

anowlcalledjosh:
*just because a PR is taking a long time doesn't mean it is no longer relevant*

*though sometimes I do wonder why this one is taking so long*

anowlcalledjosh:
bump

mlinksva:
Finally got around to making this PR consistent with the normalization improvements introduced in #342

lib/licensee/content_helper.rb
```diff
@@ -24,6 +24,9 @@ module ContentHelper
       developed_by:        /#{START_REGEX}developed by:.*?\n\n/im,
       quote_begin:         /[`'"‘“]/,
       quote_end:           /[`'"’”]/,
+      cc_legal_code:       /^\s*Creative Commons Legal Code\s*$/i,
+      cc0_info:            /For more information, please see\s*\S+zero\S+/im,
+      cc0_disclaimer:      /CREATIVE COMMONS CORPORATION.*?\n\n/im,
```
benbalter: Any concern that this might be over inclusive? Anything after any text that reads `CREATIVE COMMONS CORPORATION` would be stripped, if I'm reading correctly? Wondering if we can't match the exact normalized text?

mlinksva: Anything up to a new paragraph. I'm not particularly concerned about it being over inclusive. The whole optional paragraph is long, but if you'd prefer I could certainly change to look for a match agains the whole (normalized) paragraph:

> CREATIVE COMMONS CORPORATION IS NOT A LAW FIRM AND DOES NOT PROVIDE LEGAL SERVICES. DISTRIBUTION OF THIS DOCUMENT DOES NOT CREATE AN ATTORNEY-CLIENT RELATIONSHIP. CREATIVE COMMONS PROVIDES THIS INFORMATION ON AN "AS-IS" BASIS. CREATIVE COMMONS MAKES NO WARRANTIES REGARDING THE USE OF THIS DOCUMENT OR THE INFORMATION OR WORKS PROVIDED HEREUNDER, AND DISCLAIMS LIABILITY FOR DAMAGES RESULTING FROM THE USE OF THIS DOCUMENT OR THE INFORMATION OR WORKS PROVIDED HEREUNDER.

PR #254, opened 2017-12-29
WIP: SPDX template matching

benbalter:
An early attempt to match licenses based on SPDX templates with fields and optional sections.


lib/licensee/license_template.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    include ContentHelper
+    attr_reader :spdx_id
+
+    VAR_REGEX = /var;name=\"(?<name>.+?)\"/
+    ORIGINAL_REGEX = /original=\"(?<original>.+?)\"/
+    MATCH_REGEX = /match=\"(?<match>.+?)\"/
+    FIELD_REGEX = /(?<field><<#{VAR_REGEX};#{ORIGINAL_REGEX};#{MATCH_REGEX}>>)/
+    OPTIONAL_REGEX = /<<beginoptional>>(?:\\ )?(.*?)<<endoptional>>(?:\\ )?\s*/
```
wking: You seem to already have these stubbed out, but these forms are [very weakly specified][1].  For example, the spec does not cover how to escape a semicolon occurring in the original text.  For a much saner upstream format, you can use license-list-XML (e.g. [`ISC.xml`][2]), which leans on XML for all the hard parts.  The downside is that the XML format is [currently unstable][3], but if you vendor the XML for the licenses you need that shouldn't be a problem.

If you stick with your regexp approach and the license-list-data content, I'd recommend using `[^;]*` in your regexps instead of `.+?`.  Although I don't really understand `.+?`, so maybe it's fine.

[1]: https://spdx.org/spdx-specification-21-web-version#h.2mjng0vqrghe
[2]: https://github.com/spdx/license-list-XML/blob/cdb5b301a99c35d2aa83daa47c6ac4eba3a810df/src/ISC.xml
[3]: https://github.com/spdx/license-list-XML/tree/cdb5b301a99c35d2aa83daa47c6ac4eba3a810df#this-repository-is-unstable

mlinksva: @wking I tend to agree; the templates make me want to run away screaming, but I'm not entirely sure that's a rational response. :smile: I'd love to see guidance from SPDX that the XML is what third parties should consume if they want to leverage SPDX's careful semantic markup. Is there any hint that is even a goal?

wking: > I'd love to see guidance from SPDX that the XML is what third parties should consume if they want to leverage SPDX's careful semantic markup.

SPDX is currently warning folks away from the XML toward the old-style templates, e.g. [here][1].

> Is there any hint that is even a goal?

*I* would like the XML to be the preferred interface, since it avoids the under-specification I mention above.  The XML has its own under-specification issues (spdx/license-list-XML#495), but they're at the “what does this mean” level.  That's better than the old-style template's “there's no way to write optional text that contains semcolons” issues.

I'm not aware of folks with SPDX commit access who share my view.  My impression is that the XML will eventually become the recommended format, but it will take a bit for some schema changes (e.g. spdx/license-list-XML#452) to settle.  At the current pace, I'd estimate a year before the XML format stabilizes and then maybe another six months before folks are willing to commit a formal recommendation to consume the XML.

I don't expect the schema changes to be particularly troublesome to keep up with for consumers that pin their license-list-XML dependencies to released versions, so if it was me, I'd make the switch now.  But sticking with the old format and assuming the escaping issues don't come up will *probably* work.  And it will save you a bit of dev time while the XML format settles down.

[1]: https://github.com/spdx/license-list-XML/tree/48a9dda95d6baccaa4eb646557cdc4a7cb45e748#this-repository-is-unstable

wking:
Bump ;).

wking:
bump

wking:
bump

ale5000-git:
+1

benbalter:
This is not currently a priority for me. If anyone else would like to pick up this branch where I left off with a new PR, I'd be glad to review the implementation. Thanks for understanding.

lib/licensee/license_template.rb
goneall: @wking @mlinksva I just ran across this thread.  I think there is general support for XML in the next major release of the license list, but it is likely 8 to 12 months away.  I don't know of anyone who disagrees supporting the XML format.  This could be accelerated there is a strong demand.  Suggest this be raised with the SPDX legal team and/or SPDX tech team if you would like to accelerate the standardization of the XML format.

PR #257, opened 2018-01-03
Add support for matching licenses in Cargo.toml

andrew:
Cargo is the package manager for Rust, manifest files are written in toml

### DIFF
--- a/lib/licensee/matchers/cargo.rb
+++ b/lib/licensee/matchers/cargo.rb
@@ -1,0 +1,16 @@
+module Licensee
+  module Matchers
+    class Cargo < Licensee::Matchers::Package
+      LICENSE_REGEX = /
+        ^\s*[\'\"]?license[\'\"]?\s*=\s*[\'\"]([a-z\-0-9\._]+)[\'\"]\s*
+      /ix
+
+      private
+
+      def license_property
+        match = @file.content.match LICENSE_REGEX
+        match[1].downcase if match && match[1]
+      end
+    end
+  end
+end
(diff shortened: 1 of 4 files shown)


benbalter:
@andrew Wow. Thanks for this. Beautifully written. Tried to review, but not a single comment. 🎉

andrew:
@benbalter awesome, in that case I'll send a few more prs over for other package managers too, we're looking at using licensee in https://libraries.io this year.

benbalter:
@andrew That'd be awesome.  More than glad to review PRs. Feedback (e.g. on the API) also welcome. Just invited you to be a collaborator in case the branching model is easier.

andrew:
Awesome, thanks!

lib/licensee/matchers/cargo.rb
```diff
@@ -0,0 +1,16 @@
+module Licensee
+  module Matchers
+    class Cargo < Licensee::Matchers::Package
+      LICENSE_REGEX = /
+        ^\s*[\'\"]?license[\'\"]?\s*=\s*[\'\"]([a-z\-0-9\._]+)[\'\"]\s*
```
wking: You may also want to allow internal parens.  They are not yet supported by crates.io (withoutboats/license-exprs#3, and I'm documenting that limitation in rust-lang/cargo#4898), but hopefully they'll be supported at some point.  And I don't see a problem with allowing them in this regex now.

wking: > You may also want to allow internal parens…

I've filed support for this in #261, although Licensee will still return `other` for those cases.

PR #258, opened 2018-01-04
multiple licenses in Cargo.toml obtains other rather than nil

mlinksva:
Analogous to #240

http://doc.crates.io/manifest.html#package-metadata says `/` can separate multiple licenses (unfortunately underspecified in my view; a SPDX license expression instead, as used by NPM, makes it unambiguous whether `OR` or `AND` is intended in multiple license situations).

Also remove `_` which is not used in license ids, add `+` which is.

### DIFF
--- a/lib/licensee/matchers/cargo.rb
+++ b/lib/licensee/matchers/cargo.rb
@@ -2,7 +2,7 @@ module Licensee
   module Matchers
     class Cargo < Licensee::Matchers::Package
       LICENSE_REGEX = /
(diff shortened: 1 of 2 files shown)


mlinksva:
> unfortunately underspecified

I should've searched; this is a long-known issue https://github.com/rust-lang/cargo/issues/2039 https://github.com/rust-lang/cargo/issues/2174 seems no opposition to fixing, just nobody has done yet.

PR #259, opened 2018-01-04
Allow detection of license in elm-package.json

andrew:
Adds support detecting licenses from the elm package manager https://github.com/elm-lang/elm-package

Uses the `NpmBower` matcher as the format is almost identical, I didn't think it made sense to rename the matcher to `NpmBowerElm` as there are a few other package managers that use json files which would probably be able to use it too and the class name would get kinda ridiculous after a while!

I wasn't sure what the correct score would be, does 0.7 make sense or should it be higher?

### DIFF
--- a/lib/licensee/project_files/package_manager_file.rb
+++ b/lib/licensee/project_files/package_manager_file.rb
@@ -17,12 +17,13 @@ module Licensee
       }.freeze
 
       FILENAMES_SCORES = {
-        'package.json' => 1.0,
-        'LICENSE.spdx' => 1.0,
-        'Cargo.toml'   => 1.0,
-        'DESCRIPTION'  => 0.9,
-        'dist.ini'     => 0.8,
-        'bower.json'   => 0.75
+        'package.json'     => 1.0,
+        'LICENSE.spdx'     => 1.0,
+        'Cargo.toml'       => 1.0,
+        'DESCRIPTION'      => 0.9,
+        'dist.ini'         => 0.8,
+        'bower.json'       => 0.75,
+        'elm-package.json' => 0.7
       }.freeze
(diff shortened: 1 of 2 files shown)


benbalter:
> I wasn't sure what the correct score would be, does 0.7 make sense or should it be higher?

👍 The absolute value doesn't really matter... it was more a means to prioritize files prior to multiple license support, and I suspect it may be able to be dropped entirely in the next major version.

andrew:
Weird test failure on travis, can someone give it a retry?

wking:
> Adds support detecting licenses from the elm package manager https://github.com/elm-lang/elm-package

The Elm format seems to have an under-specified value issue (similar [to Haxe][1]).  The closest docs seem to [be][2]:

> The recommended license is BSD3, but of course, you can use whatever license you want.

which doesn't provide much clarity about other values (are SPDX short identifiers legal?  Are SPDX license expressions legal?  What does `BSD3` mean?).

Looking at that repository, it seems that [their `BSD3`][3] is [the SPDX `BSD-3-Clause`][4] based on [their text][5].  But I'd rather have that mapping (at least from `BSD3` → their recommended text) clearly documented before we add local code for it.

[1]: https://github.com/benbalter/licensee/pull/262#issuecomment-356359520
[2]: https://github.com/elm-lang/elm-package/tree/0.18.0#preparing-for-publication
[3]: https://github.com/elm-lang/elm-package/blob/0.18.0/elm-package.cabal#L13
[4]: https://github.com/spdx/license-list-XML/blob/v3.0/src/BSD-3-Clause.xml
[5]: https://github.com/elm-lang/elm-package/blob/0.18.0/LICENSE

PR #262, opened 2018-01-09
Allow detection of license in haxelib.json

andrew:
Adds support detecting licenses from the Haxe package manager, docs here: https://haxe.org/manual/haxelib-json.html

Very similar to https://github.com/benbalter/licensee/pull/259

### DIFF
--- a/lib/licensee/project_files/package_manager_file.rb
+++ b/lib/licensee/project_files/package_manager_file.rb
@@ -23,7 +23,8 @@ module Licensee
         'DESCRIPTION'      => 0.9,
         'dist.ini'         => 0.8,
         'bower.json'       => 0.75,
-        'elm-package.json' => 0.7
+        'elm-package.json' => 0.7,
+        'haxelib.json'     => 0.7
       }.freeze
 
       def possible_matchers
--- a/spec/licensee/project_files/package_info_spec.rb
+++ b/spec/licensee/project_files/package_info_spec.rb
@@ -13,6 +13,7 @@ RSpec.describe Licensee::ProjectFiles::PackageManagerFile do
       'dist.ini'         => 0.8,
       'bower.json'       => 0.75,
       'elm-package.json' => 0.70,
+      'haxelib.json'     => 0.70,
       'README.md'        => 0.0
     }.each do |filename, expected_score|
       context "a file named #{filename}" do


benbalter:
Two "nice to haves" to think about as we expand to additional package managers:

1. Listing the supported package managers in the documentation

2. Fixtures for each format to serve as integration tests in case we change the regex (and they differ slightly between formats).

andrew:
@benbalter shall I add the docs to the existing section in https://github.com/benbalter/licensee/blob/master/docs/customizing.md#matching-package-manager-metadata

wking:
On Tue, Jan 09, 2018 at 02:27:43PM +0000, Andrew Nesbitt wrote:
> Adds support detecting licenses from the Haxe package manager, docs
> here: https://haxe.org/manual/haxelib-json.html

These docs are troubling, with:

  license: The license under which the library is released. Can be
  GPL, LGPL, BSD, Public (for Public Domain) or MIT.

It's not clear to me if/how GPL and LGPL are versioned.  With SPDX
identifiers from before license-list 3.0, are they GPL-1.0+ (or, for
LGPL, LGPL-2.0+) under the [1]:

  If the Program does not specify a version number of the license, you
  may choose any version ever published by the Free Software
  Foundation.

clause (the LGPL's analog is [2])?

What does ‘Public’ get translated to?  Neither the SPDX nor Licensee
seem to have a label for this.

Which BSD (Licensee has BSD-2-Clause, BSD-3-Clause, and
BSD-3-Clause-Clear.  SPDX's license-list 3.0 has many:

  license-list-XML $ git ls-tree --name-only v3.0 -- src/ | grep BSD
  src/0BSD.xml
  src/BSD-1-Clause.xml
  src/BSD-2-Clause-FreeBSD.xml
  src/BSD-2-Clause-NetBSD.xml
  src/BSD-2-Clause-Patent.xml
  src/BSD-2-Clause.xml
  src/BSD-3-Clause-Attribution.xml
  src/BSD-3-Clause-Clear.xml
  src/BSD-3-Clause-LBNL.xml
  src/BSD-3-Clause-No-Nuclear-License-2014.xml
  src/BSD-3-Clause-No-Nuclear-License.xml
  src/BSD-3-Clause-No-Nuclear-Warranty.xml
  src/BSD-3-Clause.xml
  src/BSD-4-Clause-UC.xml
  src/BSD-4-Clause.xml
  src/BSD-Protection.xml
  src/BSD-Source-Code.xml

The only haxelib.json license that looks like an SPDX short identifier
is ‘MIT’.  Perhaps the others can be salvaged with a translation
table, although it would be good to have that table backed by
haxelib.json docs to increase the chances that haxelib.json authors
are thinking about the same licenses when they write the file.

[1]: https://github.com/spdx/license-list-XML/blob/v3.0/src/GPL-1.0.xml#L245-L247
[2]: https://github.com/spdx/license-list-XML/blob/v3.0/src/LGPL-2.0.xml#L531-L533

mlinksva:
I don't think a translation table is posssible; the haxe documentation is just too non-specific. The result is `mit` is the only license that can be detected. Anything else, if there's a license field, will obtain `other`. That seems OK to me. It'll automatically start detecting more when/if haxe users start using SPDX identifiers for other licenses. :smile:

wking:
> It'll automatically start detecting more when/if haxe users start using SPDX identifiers for other licenses. 😄

With this approach, there will be a risk of false-matches if the Haxe folks add other identifiers which happen to match SPDX license IDs but which have different semantics.  This seems unlikely, but might be possible in the BSD family.  I don't have a problem with Licensee taking that risk if they feel it is manageable.

There's also a chance that the Haxe folks add well-specified values that differ from the SPDX short identifiers and/or license expressions.  But we can always update to address those if/when that happens.

PR #263, opened 2018-01-09
Test against ruby 2.5.0

andrew:
Adding a description to make @request-info bot happy, the title is pretty self-explanatory


benbalter:
Sorry I missed this when it was originally opened, but thanks for putting up with my robots run amok nonetheless.

PR #270, opened 2018-02-05
Add shebang to bootstrap script

tchajed:
Add `#!/bin/bash` so the script can be executed directly.


sschuberth:
Actually, that should have been just `sh` instead of `bash` as there's nothing bash-specific in this script, and for consistency with other scripts, see https://github.com/benbalter/licensee/pull/272.

PR #272, opened 2018-02-12
Consistently use bin/sh shebang for shell scripts

sschuberth:


benbalter:
Nice catch. Thanks for the PR.

PR #277, opened 2018-03-12
document how to check against expected license

mlinksva:
Inspired by https://github.com/benbalter/licensee/issues/276#issuecomment-372448925


aaronfranke:
@mlinksva Is there any reason why the license would be detected [correctly in one repository](https://github.com/aaronfranke/godot-demo-projects) and [not detected correctly in another](https://github.com/aaronfranke/godot) if the licenses in each repository have the exact same name and contents? The MD5 hash is the same for both.

mlinksva:
Most likely because one of the repositories you cite has multiple files named copyright/license or similar. Unless they all are detected as having the same license, licensee's overall evaluation will be NOASSERTION -- it can't be certain of which license applies to the repo overall, and licensee values avoiding false positives.

PR #278, opened 2018-03-14
correct bad information about license notice in source files

ian-kelling:
https://www.gnu.org/licenses/identify-licenses-clearly.html:
"developers could release a program under GNU GPL
version 1 only, or release it under GPL version 1 or any later
version. The way developers state their choice is in the license notice
that goes at the start of each source file. That's where the GPL says
the decision is stated."

https://spdx.org/using-spdx#identifiers:
"The SPDX Work-group proposes to use SPDX license identifiers to
indicate the license at the file level."

### DIFF
--- a/docs/what-we-look-at.md
+++ b/docs/what-we-look-at.md
@@ -37,7 +37,7 @@ Because it's not legally binding. A license is a legal contract. You give up cer
 
 Most popular licenses today *require* that the license itself be distributed along side the software. Simply putting the letters "MIT" or "GPL" in a configuration file doesn't really meet that requirement. Those files are designed to be read by computers (who can't enter into contracts), not humans (who can). It's great metadata, but that's about it.
 
-From a practical standpoint, every language has its own package manager (some even have multiple). That means that if you want to detect the license of an arbitrary project, you'll have to implement [100s](https://github.com/github/linguist/tree/master/samples) of package-manager-specific detection strategies. The LICENSE file is a platform-agnostic and unambiguous way to communicate license intention.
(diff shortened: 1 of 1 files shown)


ale5000-git:
I think it is asking to detect whether it is "GNU General Public License v1.0 only" or "GNU General Public License v1.0 or later".

mlinksva:
In c311866 I generalized @ian-kelling's edit, and also brought up to date: licensee for a long time has optionally looked at some of the things this file says it doesn't.

ale5000-git:
This isn't true:
`It's a lot of work, as there's no standardized, cross-platform way to describe a _project's_ license within a source file comment`

There is a standard way to describe a project's license within a source file comment: https://spdx.org/using-spdx-license-identifier

mlinksva:
Yes it is true. I emphasized project for exactly the reason you cite, @ale5000-git.

benbalter:
@ale5000-git do you have an easy way to detect code comments across languages? If so, we could look into implementing this.

ale5000-git:
@benbalter 
It could simply look for `SPDX-License-Identifier: ` text in source code comments.

Example for single license:
`SPDX-License-Identifier: GPL-2.0`
Example for multiple license:
`SPDX-License-Identifier: (LGPL-2.1 AND BSD-2-CLAUSE)`

The comment is usually in a separate line from other comments (but I'm not sure if it is the rule).
But it can be easily detected since the license identifier doesn't contain spaces so a space or newline end the identifier, if there are multiple licenses then there are also the brackets to help detecting the end.
You can find the exact specs in this pdf: https://spdx.org/sites/cpstandard/files/pages/files/using_spdx_license_list_short_identifiers.pdf

PR #281, opened 2018-03-26
autoload octokit as it won't be used unless project is remote repo

### LINKED ISSUE #268: Lazy load octokit when required
To keep it out of memory if remote repos aren't used.

mlinksva:
Which will never be the case for some deployments. Fixes #268 

Seems to work, see below. I'm not of a good way to write a test since whether it is loaded would depend on whether any test accessing a remote repo had been run.

```
$ irb
irb(main):001:0> require 'licensee'
=> true
irb(main):002:0> defined?(Octokit)
=> nil
irb(main):003:0> Licensee.project('.')
=> #<Licensee::Projects::GitProject:0x00007fea220e5c90 @repository=#<Rugged::Repository:70321785220620 {path: "/Users/mlinksva/G/licensee/.git/"}>, @revision=nil, @detect_packages=false, @detect_readme=false>
irb(main):004:0> defined?(Octokit)
=> nil
irb(main):005:0> Licensee.project('https://github.com/benbalter/licensee')
=> #<Licensee::Projects::GitHubProject:0x00007fea2240e7a0 @repo="benbalter/licensee", @detect_packages=false, @detect_readme=false>
irb(main):006:0> defined?(Octokit)
=> "constant"
```

### DIFF
--- a/lib/licensee/projects/github_project.rb
+++ b/lib/licensee/projects/github_project.rb
@@ -5,7 +5,7 @@
 # Only the root directory of a repository will be scanned because every
 # `#load_file(..)` call incurs a separate API request.
 
-require 'octokit'
+autoload :Octokit, 'octokit'
 
 module Licensee
   module Projects


mlinksva:
> Huh. I had no idea you could do that.

You must have had the idea and forgotten because https://github.com/benbalter/licensee/commit/c98761e2e74d0787114391e0f57ce926d40422b5 😄  which I suppose must've reduced memory upon requiring licensee a bit (none of licensee's own components being autoloaded there are large).

I'm not sure the module scoping of autoloads seen in commit above -- and seems to be a common pattern -- accomplishes, couldn't find an explanation with quick web search.

benbalter:
I knew you could use it for subclasses/modules, but didn't realize you could use it for external Gems. It's in the loadpath, and seems to work, so 👍

PR #290, opened 2018-04-08
fix #287: allow COPYING.LESSER or LICENSE.LESSER

### LINKED ISSUE #287: Cannot detect the LGPL license
Hi,
it would be nice to detect `COPYING.LESSER` and `LICENSE.LESSER` file for the LGPL license.

Example project: https://github.com/ale5000-git/pycompatlayer

The official raccomandation on https://www.gnu.org/licenses/gpl-howto.en.html is:
> - Add a COPYING file with a copy of the GNU GPL or GNU AGPL.
> - Also add a COPYING.LESSER file with a copy of the GNU LGPL, if you use that.
> 
> You should also include a copy of the license itself somewhere in the distribution of your program. All programs, whether they are released under the GPL or LGPL, should include the text version of the GPL. In GNU programs we conventionally put the license in a file called COPYING.
> 
> Please note that, since the LGPL is a set of additional permissions on top of the GPL, it's crucial to include both licenses so users have all the materials they need to understand their rights.

1138-4EB:


1138-4EB:
Closed because of [287#issuecomment-379788335](https://github.com/benbalter/licensee/issues/287#issuecomment-379788335).

PR #293, opened 2018-04-19
Normalize http: to https:

mlinksva:
Reflectng consensus in https://github.com/spdx/license-list-XML/issues/633

Also allowing updating of some vendored licenses to https reflecting what some license stewards have done eg https://github.com/github/choosealicense.com/pull/543#issue-147444584 without any risk of false negatives.

### DIFF
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -82,6 +82,7 @@ module Licensee
         string, _partition, _instructions = string.partition(END_OF_TERMS_REGEX)
         string = strip_markup(string)
         string = normalize_quotes(string)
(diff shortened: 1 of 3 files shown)


lib/licensee/content_helper.rb
```diff
@@ -178,5 +179,9 @@ def strip(string, regex)
     def normalize_quotes(string)
       string.gsub(/\s'([\w -]*?\w)'/, ' "\1"')
     end
+
+    def normalize_https(string)
+      string.gsub(/http:/, 'https:')
```
benbalter: I believe you'll want a `/i` flag on this to make it case insensitive.

PR #294, opened 2018-04-21
Update vendor/choosealicense for UPL 1.0

brennanfee:
Updated the vendor files for ChooseALicense.com to get updated versions
of CC-BY licenses and bring in the new UPL 1.0 license.


brennanfee:
Ok, so it's not as simple as just adding the files.  I'll work on the other changes and re-submit.

PR #314, opened 2018-07-10
Autoload rugged

jonabc:
Similar to https://github.com/benbalter/licensee/pull/281, I noticed that `rugged` was always being loaded, but only used from `Licensee::Projects::GitProject`.  This PR changes `licensee` to only load `rugged` when the gem would be used.

```
➜  licensee git:(autoload-rugged) script/console
[1] pry(main)> defined?(Rugged)
=> nil
[2] pry(main)> Licensee.project(".")
=> #<Licensee::Projects::GitProject:0x007fb9d1977088
 @detect_packages=false,
 @detect_readme=false,
 @repository=#<Rugged::Repository:70218030445920 {path: "/Users/jonabc/jonabc/licensee/.git/"}>,
 @revision=nil>
[3] pry(main)> defined?(Rugged)
=> "constant"
```


jonabc:
@benbalter @mlinksva I wonder why linting errors are happening all of a sudden.  Should I fix those in this PR?

mlinksva:
Probably a new version of rubocop introduced new rules. Happens fairly often. I usually try to fix in the PR they first occur in eg last two commits in https://github.com/benbalter/licensee/pull/307.

jonabc:
Not sure the best way to get rubocop 👍 .  The complaint about the target ruby version in particular.  If I set the target ruby version (2.1) in `.rubocop.yml` to match the gemspec as it suggests I get the following:

```
Error: Unsupported Ruby version 2.1 found in `TargetRubyVersion` parameter (in .rubocop.yml). 2.1-compatible analysis was dropped after version 0.58.
Supported versions: 2.2, 2.3, 2.4, 2.5
```

Seems like rubocop might need to be locked to < 0.58 until the minimum ruby version is bumped to 2.2

mlinksva:
I guess the options are:

1. pin to last version of rubocop that supports 2.1
2. disable `Gemspec/RequiredRubyVersion` in `rubocop.yml`
3. increment `gem.required_ruby_version = '>= 2.1'` and drop `2.1` from `.travis.yml`

https://github.com/rubocop-hq/rubocop/issues/5260 seems to recommend against (2).

(3) seems OK to me since even Ruby 2.2 is recently unsupported https://www.ruby-lang.org/en/news/2018/06/20/support-of-ruby-2-2-has-ended/

benbalter:
Went with option 3 and merged in master.

catandsteve:
jonabc:autoload-rugged

lib/licensee/projects/git_project.rb
```diff
@@ -5,6 +5,9 @@
 # Project files for this project type will contain the following keys:
 #  :name - the file's path relative to the repo root
 #  :oid  - the file's OID
+
+autoload :Rugged, 'rugged'
+
 module Licensee
   module Projects
     class GitProject < Licensee::Projects::Project
```
catandsteve, line 13: @@ -1,7 +1,6 @@
require_relative 'licensee/version'
require 'forwardable'
require 'pathname'
require 'rugged'
require 'yaml'

module Licensee
 3  lib/licensee/projects/git_project.rb 
Viewed
@@ -5,6 +5,9 @@
# Project files for this project type will contain the following keys:
#  :name - the file's path relative to the repo root
#  :oid  - the file's OID

autoload :Rugged, 'rugged'

module Licensee
  module Projects
    class GitProject < Licensee::Projects::Project

PR #322, opened 2018-08-06
manufacture source meta from spdx id

mlinksva:
Because explicit source meta isn't really needed and I'd like to remove in/from https://github.com/github/choosealicense.com/pull/601 and conceivably some licensee user could be depending on `license.meta["source"]` returning a URL.


lib/licensee/license_meta.rb
```diff
@@ -53,5 +58,9 @@ def [](key)
       key = 'spdx_id' if key == 'spdx-id'
       super(key)
     end
+
+    def spdx_license_url(spdx_id)
+      return if spdx_id.nil?
```
benbalter: Should this return something if it's not null?

mlinksva: whoops, that's not used, must be leftover from experiment with different approach. will rm.

PR #336, opened 2018-10-09
Return human readable name for pseudo licenses

benbalter:
In https://github.com/benbalter/licensee/pull/297 we set the proper `spdx_id` for the `other` and `no-license` pseudo licenses. In https://github.com/benbalter/licensee/commit/7166a9cf3bf424e19f2998e1ba947c18cb9926a6, the `License#name` logic changed slightly to look to the SPDX ID, instead of the key, when no title was present in the metadata.

Both changes were valid changes, and in the case of most licenses, would result in a better non-title fallback, but combined, changed the intended-to-be-human-readable`License#name` for pseudo licenses from the capitalized key to the raw SPDX ID (e.g., `NOASSERTION` instead of `No license`).

This PR updates the logic for pseudo licenses to return the humanized key, rather than the raw SPDX ID ensuring `Licensee::License.find('other').name` returns `Other` rather than `NOASSERTION`, which is more human readable (and the old behavior).

Additionally in tracking this bug down, I noticed we (I) switched between `pseudo` and `psuedo` (and in some cases `psudo`), meaning arguments weren't always properly passed around. This PR also correct the spelling of the argument in a backwards compatible way such that `psuedo` will be transformed to `pseudo` if `psuedo` is not otherwise passed.

### DIFF
--- a/spec/licensee/license_spec.rb
+++ b/spec/licensee/license_spec.rb
@@ -44,6 +44,7 @@ RSpec.describe Licensee::License do
       it 'includes featured licenses' do
         expect(licenses).to include(mit)
         expect(licenses).to_not include(cc_by)
+        expect(licenses).to_not include(other)
       end
     end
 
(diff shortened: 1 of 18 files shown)


spec/licensee/license_spec.rb
```diff
@@ -201,7 +222,8 @@
     end
 
     it 'uses the default name when none exists' do
-      expect(other.name).to eql('NOASSERTION')
+      expect(other.name).to eql('Other')
```
LindseyB: ✨

PR #342, opened 2018-10-25
Better content normalization

benbalter:
This PR ports the normalizations from #254 into its own PR so they can be merged parallel to the SPDX work. Specifically:

* Normalizes bullets
* Normalizes urls
* Normalizes ampersands
* Normalizes dashes
* Normalizes SPDX "Varietal" words
* Better markup stripping
* Better title regex
* Better Copyright regex
* General normalization abstraction

TODO: 

- [x] Tests for ☝️


lib/licensee/content_helper.rb
```diff
@@ -4,17 +4,87 @@
 module Licensee
   module ContentHelper
     DIGEST = Digest::SHA1
+    START_REGEX = /\A\s*/
```
mlinksva: Just curious, why do this and the next two remain outside the hash?

benbalter: Missed a few. Fixed. Good catch,

PR #357, opened 2019-01-12
Fixing cabal licenses handling

### LINKED ISSUE #356: licensee GPL-3.0 specifier does not match haskell's cabal format
### Describe the bug

I maintain a haskell project which is licensed using GPL-3.0. If I specify `GPL-3.0`  in the `.cabal` file, licensee detects it properly. However, If I run `cabal check`, it warns that the specifier is wrong:  

```
These warnings may cause trouble when distributing the package:
* 'license: GPL-3.0' is not a known version of that license. The known
versions are 2, 3. If this is not a mistake and you think it should be a known
version then please file a ticket.
```

This is because cabal expects `GPL-3` - no minor version. If I change the version from `GPL-3.0` to `GPL-3` and commit changes, `cabal check` works....

```
cabal check
No errors or warnings could be found in the package.
```

...but licensee does not: 

```
licensee 
License:        NOASSERTION
Matched files:  LICENSE, mulang.cabal
LICENSE:
  Content hash:  fb4942f66306942d6b76feff8169f4d138fae884
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       GPL-3.0
mulang.cabal:
  Confidence:  90.00%
  Matcher:     Licensee::Matchers::Cabal
  License:     NOASSERTION
```

1. Go to 'https://github.com/mumuki/mulang'
2. Edit and commit `.cabal` file to use `GPL-3`
3. Run licensee and verify that it is not recognizing the license, although that is the expected cabal format 

### Expected behavior

licensee should accept `GPL-3` and `GPL-2` specifiers in cabal files, and treat them exactly as `GPL-3.0` and `GPL-2.0`.

flbulgarelli:
Fixes #356

This PR only affects some licenses: 

* most GPL and GPL-derivative license
* BSD licenses

The other already supported licenses where already working. Also, unrecognized cabal licenses are preserved without changes. 

Finally, there are some other cabal licenses that are not explicitly handled, as you can check with an invalid cabal file: 

``
$ cabal check
Warning: The following warnings are likely to affect your build negatively:
Warning: 'license: VGPL-4.0' is not a recognised license. The known licenses
are: GPL, GPL-2, GPL-3, LGPL, LGPL-2.1, LGPL-3, AGPL, AGPL-3, BSD2, BSD3, MIT,
ISC, MPL-2.0, Apache, Apache-2.0, PublicDomain, AllRightsReserved,
OtherLicense
```

I did not add explicit support for them just because different reasons: 

* `GPL`, `APL`, `LGPL` without version numbers are ambiguous 
* `PublicDomain`, `OtherLicense`, `AllRightsReserved` are not actually licenses, as far as I understand.

### DIFF
--- a/lib/licensee/matchers/cabal.rb
+++ b/lib/licensee/matchers/cabal.rb
@@ -4,12 +4,29 @@ module Licensee
       # While we could parse the cabal file, prefer
       # a lenient regex for speed and security. Moar parsing moar problems.
       LICENSE_REGEX = /^\s*license\s*\:\s*([a-z\-0-9\.]+)\s*$/ix.freeze
+      LICENSE_CONVERSIONS = {
+        'GPL-2'      => 'GPL-2.0',
+        'GPL-3'      => 'GPL-3.0',
+        'LGPL-2.1'   => 'LGPL-2.1',
+        'LGPL-3'     => 'LGPL-3.0',
+        'AGPL-3'     => 'AGPL-3.0',
+        'BSD2'       => 'BSD-2-Clause',
+        'BSD3'       => 'BSD-3-Clause',
+        'MIT'        => 'MIT',
+        'ISC'        => 'ISC',
+        'MPL-2.0'    => 'MPL-2.0',
+        'Apache-2.0' => 'Apache-2.0'
+      }.freeze
 
       private
 
       def license_property
         match = @file.content.match LICENSE_REGEX
-        match[1].downcase if match && match[1]
+        spdx_name(match[1]).downcase if match && match[1]
+      end
+
+      def spdx_name(cabal_name)
+        LICENSE_CONVERSIONS[cabal_name] || cabal_name
       end
     end
   end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lib/licensee/matchers/cabal.rb [hunk 1]
```diff
@@ -4,12 +4,29 @@ class Cabal < Licensee::Matchers::Package
...
+        'GPL-3'      => 'GPL-3.0',
+        'LGPL-2.1'   => 'LGPL-2.1',
+        'LGPL-3'     => 'LGPL-3.0',
+        'AGPL-3'     => 'AGPL-3.0',
+        'BSD2'       => 'BSD-2-Clause',
+        'BSD3'       => 'BSD-3-Clause',
+        'MIT'        => 'MIT',
+        'ISC'        => 'ISC',
+        'MPL-2.0'    => 'MPL-2.0',
+        'Apache-2.0' => 'Apache-2.0'
```
mlinksva: I don't think the last 4 are needed as they are no-ops.

mlinksva:
👍 on not adding conversions for ambiguous or non-licenses. Licensee will find `other` for any id it doesn't know about https://github.com/benbalter/licensee/blob/16c2b2385911cb5a6af1dadeaba2ac3727ea3a96/lib/licensee/matchers/package.rb#L8-L11

Left one minor comment. Looks good to me!

lib/licensee/matchers/cabal.rb
flbulgarelli: Ooops, you are right :ok_hand:. Removing them and `LGPL-2.1` which is already in standard SPDX format.

PR #359, opened 2019-02-04
Remove duplicated line

cxong:
Remove a duplicated line in docs


mlinksva:
The removed line wasn't a duplicate, rather it shows that both `s` and `c` spellings are supported.

PR #365, opened 2019-02-16
Vendored licenses: add 0BSD, 'how' field fixes

mlinksva:
cc https://github.com/github/choosealicense.com/pull/643


benbalter:
Curious, it looks like the hashes didn't change, but the order did?

mlinksva:
Exactly. As mentioned in commit message for 50c0258e48e4dec64d9a0801d9937f5226855e2f I guess the sorting must be from some library change, but I didn't investigate further beyond looking at `JSON.pretty_generate` to see if sorting was an explicit option so we wouldn't see a change again, but I didn't see such a thing.

benbalter:
https://github.com/benbalter/licensee/commit/096e009e7c2417fa29add4dda348fa87395560eb is the culprit. I'll re-hash on `master` which should clean up the diff here.

Issue #312, opened 2018-07-02
The `licensee detect` command gives error after initializing an empty git repository


ArniDagur:
### Description of bug

`licensee detect` gives error after initializing an empty git repository

### Steps to reproduce the behavior

0.  `licensee detect` works fine with no git repository

```
[arni: LicenseeTest]$ ls
1stAncestor  LICENSE
```
```
[arni: LicenseeTest]$ licensee detect
License:        Apache-2.0
Matched files:  LICENSE
LICENSE:
  Content hash:  4579921419ac7db65117c87a46ad784f4a32bb01
  Confidence:    99.52%
  Matcher:       Licensee::Matchers::Dice
  License:       Apache-2.0
  Closest licenses:
    Apache-2.0 similarity:  99.52%
    ECL-2.0 similarity:     96.36%
    EPL-2.0 similarity:     57.20%
```
1. Initialize empty repo with `git init`
```
[arni: LicenseeTest]$ git init
Initialized empty Git repository in /home/arni/LicenseeTest/.git/
```
2. `licensee detect` gives error
```
[arni: LicenseeTest]$ licensee detect
Traceback (most recent call last):
        16: from /home/arni/.gem/ruby/2.5.0/bin/licensee:23:in `<main>'
        15: from /home/arni/.gem/ruby/2.5.0/bin/licensee:23:in `load'
        14: from /home/arni/.gem/ruby/2.5.0/gems/licensee-9.9.1/bin/licensee:37:in `<top (required)>'
        13: from /home/arni/.gem/ruby/2.5.0/gems/thor-0.20.0/lib/thor/base.rb:466:in `start'
        12: from /home/arni/.gem/ruby/2.5.0/gems/thor-0.20.0/lib/thor.rb:387:in `dispatch'
        11: from /home/arni/.gem/ruby/2.5.0/gems/thor-0.20.0/lib/thor/invocation.rb:126:in `invoke_command'
        10: from /home/arni/.gem/ruby/2.5.0/gems/thor-0.20.0/lib/thor/command.rb:27:in `run'
         9: from /home/arni/.gem/ruby/2.5.0/gems/licensee-9.9.1/lib/licensee/commands/detect.rb:23:in `detect'
         8: from /home/arni/.gem/ruby/2.5.0/gems/licensee-9.9.1/lib/licensee/projects/project.rb:24:in `license'
[...]

PR #367, opened 2019-02-27
fix error on new git init

atakanyenel:
This PR fixes the issue #312  . If the repo is fresh initialized, the commit function in GitProject fails as there is no commit , to prevent this , thit PR checks for a head_unborn? in the initialize and raises InvalidRepository error, so that the project can be treated as File System project.

I wasn't sure how to write tests for this. I couldn't find any similar example of it in the specs.

EDIT: There seems to be 2 ways to decrease cognitive complexity. Either I can call commit function directly, raise `rugged::ReferenceError` and then catch it in rescue or we can move `if repo.is_a? Rug..` block to a private method.

### DIFF
--- a/lib/licensee/projects/git_project.rb
+++ b/lib/licensee/projects/git_project.rb
@@ -22,6 +22,8 @@ module Licensee
           Rugged::Repository.new(repo)
         end
 
+        raise InvalidRepository if repository.head_unborn?
+
         @revision = revision
         super(**args)
       rescue Rugged::OSError, Rugged::RepositoryError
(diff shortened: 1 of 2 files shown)


benbalter:
@atakanyenel I really appreciate this and sorry for the codeclimate hassle... something new I'm trying with this project. How does https://github.com/benbalter/licensee/pull/367/commits/64b8e640844bd0783625980776ae5200e819339f look to you? Does that resolve the error on git init for you locally?

atakanyenel:
The commit works for me. I also added a test for the new added line. The test made most sense in it's own file as `project_spec.rb` had different `before do` commands.

PR #372, opened 2019-03-07
Better detect the LGPL

### LINKED ISSUE #287: Cannot detect the LGPL license

cbarrick:
This makes two changes to better support the LGPL, from the discussion in #287.

1. License files are now allowed to have multiple extensions, e.g. `COPYING.LESSER.md`.
2. The LGPL license file is no longer required to be named `COPYING.LESSER`.

The first change has implications beyond the LGPL, but I can't think of any negative consequences.

The second change only removes code, and I don't think I need to add any new tests for this change. The old `LGPL` context in the `LicenseFile` spec is removed because it's all about handling a file specifically named `COPYING.LESSER`. The `lgpl` context in the `Project` spec ensures that the LGPL takes precedence when license files for both the GPL and LGPL are present, which I believe is a sufficient test.

Together these close #287.

### DIFF
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -8,7 +8,7 @@ module Licensee
       PREFERRED_EXT_REGEX = /\.#{Regexp.union(PREFERRED_EXT)}\z/.freeze
 
       # Regex to match any extension except .spdx or .header
-      OTHER_EXT_REGEX = %r{\.(?!spdx|header|gemspec)[^./]+\z}i.freeze
+      OTHER_EXT_REGEX = %r{(\.(?!spdx|header|gemspec)[^./]+)+\z}i.freeze
 
       # Regex to match, LICENSE, LICENCE, unlicense, etc.
       LICENSE_REGEX = /(un)?licen[sc]e/i.freeze
--- a/lib/licensee/project_files/license_file.rb
+++ b/lib/licensee/project_files/license_file.rb
@@ -68,7 +68,7 @@ module Licensee
       end
 
       def lgpl?
-        LicenseFile.lesser_gpl_score(filename) == 1 && license && license.lgpl?
+        license && license.lgpl?
       end
 
       def gpl?
@@ -85,11 +85,6 @@ module Licensee
 
       def self.name_score(filename)
         FILENAME_REGEXES.find { |regex, _| filename =~ regex }[1]
-      end
-
-      # case-insensitive block to determine if the given file is LICENSE.lesser
-      def self.lesser_gpl_score(filename)
-        filename.casecmp('copying.lesser').zero? ? 1 : 0
       end
     end
   end
--- a/spec/licensee/project_files/license_file_spec.rb
+++ b/spec/licensee/project_files/license_file_spec.rb
@@ -70,6 +70,8 @@ RSpec.describe Licensee::ProjectFiles::LicenseFile do
       'LICENCE.docs'       => 0.80,
       'copying.image'      => 0.75,
       'COPYRIGHT.go'       => 0.75,
+      'COPYING.LESSER'     => 0.75,
+      'COPYING.LESSER.md'  => 0.75,
       'LICENSE-MIT'        => 0.70,
       'LICENSE_1_0.txt'    => 0.70,
       'COPYING-GPL'        => 0.65,
--- a/spec/licensee/project_files/license_file_spec.rb
+++ b/spec/licensee/project_files/license_file_spec.rb
@@ -93,24 +93,6 @@ RSpec.describe Licensee::ProjectFiles::LicenseFile do
       end
     end
 
-    context 'LGPL scoring' do
-      {
-        'COPYING.lesser' => 1,
-        'copying.lesser' => 1,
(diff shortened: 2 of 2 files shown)


mlinksva:
I think this probably relaxes things a bit too much -- I'm more confident that the presence of a license file containing the GPL and a license file with lesser in the nam containing LGPL is following the spirit of the official documentation and actually part of an intended application of LGPL to a repo. And there should still be tests for the two file LGPL case, which at a glance it seems this PR removes.

ale5000-git:
@cbarrick: Whils you are at it please test also `COPYING.LESSER.rst` (reStructuredText) and `LICENSE.LESSER.rst`.

cbarrick:
I believe the tests for the `Project` class check for the two file case. I think I only removed tests that worry about the file name.

Requiring that the LGPL be distributed with "lesser" in the file name seems arbitrary to me. What if the user has files `LICENSE-GPL` and `LICENSE-LGPL`?

That case is reasonable and creates a false positive for the GPL in a project that is LGPL. Is there a concrete use case where this could cause a false positive for the LGPL?

Sometimes this pattern is used for dual licensing, but sicne the LGPL is strictly more permissive than the GPL, "dual licensed under the GPL and LGPL" is equivalent to "licensed under the LGPL".

mlinksva:
When there's multiple license files present, licensee finds NOASSERTION for a repo because it can't be sure what their purpose is. Dual licensing is one possibility, another is that different licenses apply to different materials in the repo, or one is for third party dependencies. In general licensee tries to be conservative, considering false positives far worse than false negatives.

In https://github.com/licensee/licensee/issues/287#issuecomment-470459587 you noticed that `COPYING.md` (containing GPL-3.0) and `COPYING.LESSER.md` (containing LGPL-3.0) causes licensee to identify GPL-3.0. That's a bug, even if we hew to only identifying LGPL-3.0 in the presence of two license files if the one containing LGPL-3.0 is specifically named `COPYING.LESSER` -- licensee should find NOASSERTION in that case.

The reason we're OK with licensee identifying LGPL-3.0 in the presence of two license files containing GPL-3.0 and LGPL-3.0 rather than NOASSERTION is because there's a specific pattern that is widely used and recommended by the license steward.

I think it is totally reasonable to still positively identify LGPL-3.0 given very minor variations on that recommendation, which we can see frequently in the wild, eg LICENSE and LICENSE.LESSER instead of COPYING and COPYING.LESSER.

Not looking for that pattern at all reduces my confidence that's what's actually going on, instead of other situations that could result in multiple license files, which brings us back to NOASSERTION.

cbarrick:
> different licenses apply to different materials in the repo, or one is for third party dependencies

Good point. I was assuming all files that match refer to the repo. My instinct was that dependency licenses wouldn't be at the project root. Under that assumption, the correct response is almost always no assertion when faced with multiple licenses.

I think the logic in this PR does this:

1. If there is one license, return that license.
2. If there are two licenses, and they are GPL and LGPL, return LGPL.
3. In any other case, return no assertion.

The false positive case here is that the author has exactly two license files in the root of their project, GPL and LGPL, where the LGPL amendments only apply to a dependency. If the author has more dependency licenses at the root, or the dependency license is stored elsewhere, this doesn't apply.

In the tradeoff, I still think the PR is an improvement. The current behavior is a false positive for LGPL users who don't follow GNU conventions. The PR's behavior is a false positive for GPL users who only have GPL and LGPL dependencies and store the dependency license in the root. I believe this new false positive is smaller in scope.

I also think the behavior is better in practical searching: It's better to not exclude an LGPL from a search that excludes GPL than to include a GPL into that search. Likewise its better to include a GPL into a search exclusive to LGPL than to exclude an LGPL from that search.

The story being that the search filtering out GPL would be used by someone seeking dependencies for a proprietary app. LGPL authors would like to appear in that search. The search filtering to LGPL would only be used for a count, and it's effected either way.

There are some things that would be nice to know:

1. How many repositories are effected by both false positives.
2. How many LGPL users only follow GNU conventions for compatibility with GitHub and Licensee.

cbarrick:
Ah, I see that the GPL-should-be-LGPL false positive only happens when the LGPL file isn't discovered at all

mlinksva:
Exactly.

I really want to avoid identifying LGPL if there's a reasonable chance GPL is the applicable license. If people are following conventions because GitHub expects conventions to be followed, that's a plus!

Issue #358, opened 2019-01-17
Incorrect yaml output for multiline attribution string


benwiley4000:
### Describe the bug

When an attribution string contains multiple lines (an example is the `accepts` module from npm, version 1.3.3), the output doesn't come out as correct yaml. For `accepts`, we get this output:

```yaml
License:        MIT
Matched files:  LICENSE
LICENSE:
  Content hash:  d64f3bb4282a97b37454b5bb96a8a264a3363dc3
  Attribution:   Copyright (c) 2014 Jonathan Ong <me@jongleberry.com>
Copyright (c) 2015 Douglas Christopher Wilson <doug@somethingdoug.com>
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       MIT
```

Yaml parsers (such as `js-yaml` from npm) don't know how to parse this and I get this error:

```
(node:18392) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): YAMLException: can not read a block mapping entry; a multiline key may not be an implicit key at line 7, column 13:
      Confidence:    100.00%
```

I consulted [this guide](https://yaml-multiline.info/) to handling multiline strings in yaml and I found this is the correct output format:

```yaml
License:        MIT
Matched files:  LICENSE
LICENSE:
  Content hash:  d64f3bb4282a97b37454b5bb96a8a264a3363dc3
  Attribution:   "Copyright (c) 2014 Jonathan Ong <me@jongleberry.com>\\nCopyright (c) 2015 Douglas Christopher Wilson <doug@somethingdoug.com>"
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       MIT
```

As you can see we surround the multiline string in double quotes, and instead of a literal newline we use `\\n` (regular `\n` is ignored). Also we would clearly need to escape any double quotes that appear in the attribution string, if there were any.

### Steps to reproduce the behavior

1. `licensee detect --remote --no-packages --no-readme jshttp/accepts`
2. You should see the same poorly formed yaml string as I printed above

### Expected behavior

Yaml I can parse appropriately

benwiley4000:
I also noticed another YAML output bug where you can get duplicate keys if the license file is called `License` .. but then I noticed there is a `--json` flag and this probably isn't meant to be read as JSON. Oops!

benbalter:
[...]

PR #381, opened 2019-05-22
Expose attribution field in JSON

benwiley4000:
This change exposes the `attribution` attribute for matched files via JSON.

Resolves the suggestion [here](https://github.com/licensee/licensee/issues/358#issuecomment-455237764) in #358.


benbalter:
@benwiley4000 It looks like the rubocop failures are unrelated. I'll try to get them fixed so we can get this PR merged. Thanks!

PR #391, opened 2019-07-17
Improved title and fill-in fields normalization

### LINKED ISSUE #386: Unable to detect BSD 4-Cause license
### Describe the bug

With the BSD 4-Clause license added to Choose a license https://github.com/github/choosealicense.com/pull/669 licensee is still unable to match the license.

1. Run ./script/cibuild using this PR https://github.com/github/choosealicense.com/pull/669

### Expected behavior

cibuild should identify the licenses as BSD 4-clause and the tests should pass.

### Additional context

The matching breaks as soon as there is a space in [fullname] or [project]. It looks like the BSD 3-Clause isn't so picky which I don't understand.  It looks like Licensee has to be updated first before the matcher will work here. They use the list of licenses from here though, so this seems like a chicken and egg problem.

mlinksva:
Fixes #386 following on to #388. 

* remove need for some per-license title regexes by expanding matching based on spdx id.
* try stripping title again after copyright, catching licenses where copyright usually (UPL-1.0) or occasionally (BSDs etc) is placed before title.
* make accounting for fill-in fields slightly more accurate, reuse existing code to extract them
* boost max_delta slightly based on presence of non-stripped fill-in fields so that appropriate licenses are actually candidates rather than not being attempted due to small character count differences when filled in fields are different lengths than field names.


spec/licensee/content_helper_spec.rb
```diff
@@ -265,9 +266,7 @@ def filename
         let(:stripped_content) { subject.content_without_title_and_version }
 
         it 'strips the title' do
-          skip "Doesn't strip ECL title" if license.key == 'ecl-2.0'
```
benbalter: :heart:

lib/licensee/license.rb
```diff
@@ -157,7 +150,14 @@ def title_regex
         Regexp.new string, 'i'
       end
 
-      parts = [title_regex, key]
+      key_regex = begin
+        string = key.sub('-', '[- ]')
+        string.sub!('.', '\.')
```
benbalter: Could potentially use `Regexp.escape` here instead.

lib/licensee/license.rb
```diff
@@ -157,7 +150,14 @@ def title_regex
...
+        string << '(?:\ licen[sc]e)?'
```
benbalter: Do we want an optional `the` prefix like above?

lib/licensee/license.rb
```diff
@@ -157,7 +150,14 @@ def title_regex
         Regexp.new string, 'i'
       end
 
-      parts = [title_regex, key]
+      key_regex = begin
```
benbalter: 🤔 are we worried that optimistically matching the key might lead to false positives? In the case of the reference matcher, if I have `# License\n I will sMITe anyone who uses this code`, wouldn't it now match as MIT? Perhaps word boundaries could help?

lib/licensee/content_helper.rb
```diff
@@ -8,7 +8,6 @@ module ContentHelper
     DIGEST = Digest::SHA1
     START_REGEX = /\A\s*/.freeze
     END_OF_TERMS_REGEX = /^[\s#*_]*end of terms and conditions\s*$/i.freeze
-    ALT_TITLE_REGEX = License::ALT_TITLE_REGEX
```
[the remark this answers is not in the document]
mlinksva: #342 removed several other constants that look to be to be the same as/peer to `ALT_TITLE_REGEX`. I thought it must be just an oversight that it wasn't removed as well.  Am I missing something obvious?

lib/licensee/license.rb
```diff
@@ -144,8 +138,7 @@ def name_without_version
     def title_regex
       return @title_regex if defined? @title_regex
 
-      title_regex = ALT_TITLE_REGEX[key]
-      title_regex ||= begin
+      title_regex = begin
```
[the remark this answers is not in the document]
mlinksva: Good point, hadn't occurred to me. I think the begin block makes the code a bit more readable, but always happy to delete lines and indentation! 😄

lib/licensee/license.rb
mlinksva: I might be missing something, but that seems more complicated to me, as some of the escaping Regexp.escape does has to be undone whether it is done before or after the line currently above.

lib/licensee/license.rb
mlinksva: I don't think so. Above is actually stripping `the` from the license name; `the` isn't part of the regex created at all. `strip_title` doesn't care about beginning of line, only that line contains a match, so `the` is irrelevant. Possibly it should be made more strict, but I don't know of any bug reports triggered by this logic.

lib/licensee/license.rb
mlinksva: We already have that bug in the reference matcher, see https://github.com/licensee/licensee/issues/392#issuecomment-515782793 ... that's in the current (0.9.11) release, but I suspect has been there for as long as the reference matcher has existed.

I agree that's a bug, but it probably needs to be fixed in the reference matcher.

In the case of MIT, this PR has zero behavior change, since the title and key each end up generating the same regex.

mlinksva:
@benbalter thanks, I responded to all of your comments, a couple have follow up questions.

alistair23:
Is this going to be merged?

mlinksva:
@benbalter re-pinging for review in hopes of closing #386 before another stalebot/bump round. 😄

lib/licensee/license.rb
```diff
@@ -144,20 +138,22 @@ def name_without_version
...
+      string.sub!(/,? version /, ' ')
+      string.sub!(/v(\d+\.\d+)/, '\1')
+      string = Regexp.escape(string)
+      string = string.sub(/\\ licen[sc]e/i, '(?:\ licen[sc]e)?')
+      string = string.sub(/\\ (\d+\\.\d+)/, ',?\s+(?:version\ |v(?:\. )?)?\1')
+      string = string.sub(/\bgnu\\ /, '(?:GNU )?')
+      title_regex = Regexp.new string, 'i'
+
+      string = key.sub('-', '[- ]')
+      string.sub!('.', '\.')
```
benbalter: Using `Regexp.escape` as the first step should be a bit more robust and consistent with the title treatment above.

benbalter:
Thanks @mlinksva. Sorry for the delay!

PR #399, opened 2019-10-11
Loosen dependency on octokit to allow later minor versions

parkr:
It seems unnecessary to lock to `octokit` `4.8.x`, so this commit allows octokit 4.x.

/cc @benbalter for review.


parkr:
Thank you so much, @benbalter!

PR #407, opened 2020-01-11
Remove duplicated line

GimunLee:
I removed a duplicated line in docs/what-we-look-at.md. 

-----
[View rendered docs/what-we-look-at.md](https://github.com/GimunLee/licensee/blob/master/docs/what-we-look-at.md)


mlinksva:
Thanks for the close (but not quite close enough 😄 ) read -- the line isn't duplicated, note S and C spellings.

PR #432, opened 2020-05-08
Add matcher for NuGet license expressions in .nuspec files

zarenner:
Adds support for matching NuGet license expressions (https://docs.microsoft.com/en-us/nuget/reference/nuspec#license).

Note that the `license` element also supports pointing to a file instead of an expression; licensee should already find the vast majority (but not all) of these though.

### DIFF
--- a/lib/licensee/matchers/nuget.rb
+++ b/lib/licensee/matchers/nuget.rb
@@ -1,0 +1,43 @@
+# frozen_string_literal: true
+
+module Licensee
+  module Matchers
+    class NuGet < Licensee::Matchers::Package
+      # While we could parse the nuspec file, prefer a lenient regex for speed and security.
+      # Moar parsing moar problems.
+      LICENSE_REGEX = %r{
+        <license\s*type\s*=\s*["']expression["']\s*>([a-z\-0-9\. +()]+)<\/license\s*>
+      }ix.freeze
+
+      LICENSE_URL_REGEX = %r{<licenseUrl>\s*(.*)\s*<\/licenseUrl>}i.freeze
+
+      NUGET_REGEX = %r{https?:\/\/licenses.nuget.org\/(.*)}i.freeze
+      OPENSOURCE_REGEX = %r{https?:\/\/(?:www\.)?opensource.org\/licenses\/(.*)}i.freeze
+      SPDX_REGEX = %r{https?:\/\/(?:www\.)?spdx.org\/licenses\/(.*?)(?:\.html|\.txt)?$}i.freeze
+      APACHE_REGEX = %r{https?:\/\/(?:www\.)?apache.org\/licenses\/(.*?)(?:\.html|\.txt)?$}i.freeze
+
+      private
+
+      def license_from_first_capture(url, pattern)
+        match = url.match(pattern)
+        match[1].downcase if match && match[1]
+      end
+
+      def license_from_url(url)
+        license_from_first_capture(url, NUGET_REGEX) ||
+          license_from_first_capture(url, OPENSOURCE_REGEX) ||
+          license_from_first_capture(url, SPDX_REGEX) ||
+          license_from_first_capture(url, APACHE_REGEX)&.gsub('license', 'apache')
+      end
+
+      def license_property
+        # Prefer the explicit <license type="expression"> element
+        match = @file.content.match LICENSE_REGEX
+        return match[1].downcase if match && match[1]
+
+        url_match = @file.content.match LICENSE_URL_REGEX
+        license_from_url(url_match[1]) if url_match && url_match[1]
+      end
+    end
+  end
+end
--- a/spec/licensee/matchers/nu_get_matcher_spec.rb
+++ b/spec/licensee/matchers/nu_get_matcher_spec.rb
@@ -1,0 +1,91 @@
+# frozen_string_literal: true
+
+RSpec.describe Licensee::Matchers::NuGet do
+  subject { described_class.new(file) }
+
+  let(:content) { '<license type="expression">mit</license>' }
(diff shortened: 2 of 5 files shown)


spec/licensee/matchers/nu_get_matcher_spec.rb
```diff
@@ -0,0 +1,57 @@
+# frozen_string_literal: true
```
zarenner, line 1: The name of this file is funny due to rubocop wanting "NuGet" -> "nu_get" - if there's a way to suppress that, I'm happy to change it. Also fine as is, your call.

mlinksva: Does it complain about `nuget`? That would seem preferable to `nu_get`, and matches the library file name.

lib/licensee/matchers/nuget.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Licensee
+  module Matchers
+    class NuGet < Licensee::Matchers::Package
+      # While we could parse the nuspec file, prefer a lenient regex for speed and security.
+      # Moar parsing moar problems.
+      LICENSE_REGEX = %r{
+        <license\s*type\s*=\s*["']expression["']\s*>([a-z\-0-9\. +()]+)<\/license\s*>
```
mlinksva, line 9: Would it be worth also matching licenseUrl where the URL likely contains a SPDX ID in the path which can be extracted?

opensource.org/licenses/{id} seems [pretty common](https://github.com/search?q=filename%3A.nuspec+licenseurl+opensource.org&type=Cod). spdx.org/licenses and licenses.nuget.org values exist but are surprisingly uncommon.

Not essential, just a thought.

spec/licensee/matchers/nu_get_matcher_spec.rb
zarenner: Yes, that's the issue:
```
spec/licensee/matchers/nuget_matcher_spec.rb:3:1: C: RSpec/FilePath: Spec path should end with licensee/matchers/nu_get*_spec.rb.
```
 I don't know my way around rubocop to suppress that.

lib/licensee/matchers/nuget.rb
zarenner: I think it would be useful for opensource.org, and perhaps others like https://www.apache.org/licenses/LICENSE-2.0.txt that I've noticed a fair number of packages use. I agree with your assessment that licenses.nuget.org isn't especially useful - most of the time it's used, there's a `<license>` expression anyways since those features were added to nuget around the same time.

But there's arguably a question of ownership between `licensee/licensee` and `github/licensed` when it comes to `licenseUrl`. My understanding is that fetching and matching the content of the URL (when it isn't well-known) might not meet the goals of `licensee` - I don't think any other project type in `licensee` currently downloads license content from the internet, whereas `licensed` already does that for some package types.

It seemed difficult to split it between the two, since `licensed` would presumably want to know if the match for `package_file` was due to a `license` or `licenseUrl`, so it could avoid downloading the license if it was well-known, but still download it in the case where `licensee` found a `license` but `licenseUrl` was a non-well-known URL. package.rb doesn't currently have a mechanism to expose something like that.

spec/licensee/matchers/nu_get_matcher_spec.rb
mlinksva: I see, it looks like rubocop wants spec naming to be derived from the tested class name in a specific way and it's possible to override, see https://www.rubydoc.info/gems/rubocop-rspec/RuboCop/Cop/RSpec/FilePath but maybe not worth overriding.

lib/licensee/matchers/nuget.rb
mlinksva: Fetching the content specified by `licenseUrl` would be out of scope for licensee IMO, but extracting license identification from well known URLs seems within scope for licensee.

Licensed _may_ wish to do a bit of duplicate work to avoid downloading well known license texts unnecessarily.

jonabc: > Fetching the content specified by licenseUrl would be out of scope for licensee IMO, but extracting license identification from well known URLs seems within scope for licensee.

I'm decidedly unopinionated on what logic should exist here vs licensed and am happy to follow opinion on what's decided here, but this generally makes sense to me 👍 

> Licensed may wish to do a bit of duplicate work to avoid downloading well known license texts unnecessarily.

👍 licensed already autogenerates text when a definitive classification is made but no license text is available from a known source.  It shouldn't be too hard to extend that to a scenario where there is 100% confidence in the license classification and a costly operation to get license text would match autogenerated text

zarenner: Cool, I'll move that logic into `licensee` then and update this PR.

spec/licensee/matchers/nu_get_matcher_spec.rb
zarenner: Going to leave it unless someone feels strongly.

lib/licensee/matchers/nuget.rb
zarenner: Added matching for `licenseUrl` to recognize licenses.nuget.org, opensource.org, spdx.org, and apache.org well-known licenses. Added relevant tests.

zarenner:
@mlinksva thanks for the review and merge. Any chance a release could be spun sometime soon, so I can consume this change in `licensed`?

mlinksva:
@benbalter ☝️ re new release.

benbalter:
Released https://github.com/licensee/licensee/releases/tag/v9.14.0. Thanks for the pull request!

PR #445, opened 2020-07-04
Modify the long text penality to fix Vim license detection

othree:
I found that Github already support Vim license https://api.github.com/licenses/vim 
But none of the three repo listed shows "Vim License"

What I found:
1. Starting from Licensee 9.14. The Vim project will not detect as Vim license. The reason is the commit https://github.com/licensee/licensee/commit/8fe6e80fd9d407fd26e3d8c0b1899fb787f1a372
    It works fine at 9.13.2

    So in this PR I update the parameter of the penalty. I believe the constant `10` is a human decide value. So after some test I use `13` instead. I also add a test case to test the Vim license.

2. The [vim-pathogen](https://github.com/tpope/vim-pathogen) can be detect as Vim license using licensee cli(9.14). Because the length of repos's name `vim-pathogen` is close to `[project]`. But on GitHub it is still not detect as "Vim License". I am not sure why this happens. Maybe GitHub only apply some matchers?

### DIFF
--- a/lib/licensee/content_helper.rb
+++ b/lib/licensee/content_helper.rb
@@ -132,7 +132,7 @@ module Licensee
       overlap = (wordset_fieldless & other.wordset).size
       total = wordset_fieldless.size + other.wordset.size -
               fields_normalized_set.size
-      (overlap * 200.0) / (total + length_delta(other) / 10)
+      (overlap * 200.0) / (total + length_delta(other) / 13)
     end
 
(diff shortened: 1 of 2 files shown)


mlinksva:
#437 should fix this when merged, for licensee. For GitHub.com, a new version of licensee would need to be deployed. For a specific repo on GitHub.com, license detection would need to be triggered (by changing the license file, or by support).

PR #448, opened 2020-07-31
Enable new cops, follow their recommendations

mlinksva:
Sort of follow up to #423 -- that approach adds new cops to .rubocop.yml to avoid warnings (which seems like make-work to me), this one just deals with new cops as they come, back to the previous behavor of rubocop, see https://docs.rubocop.org/rubocop/versioning.html#pending-cops

Almost all of the recommendations followed here are to eliminate redundant escaping in regex character classes -- only `]\^-` characters need to be escaped.


benbalter (approved):
:sparkles:

PR #455, opened 2020-08-30
universal newline encoding for project files, test for crlf file

### LINKED ISSUE #454: Regression: BSD 3-clause no longer recognized
### Describe the bug

The BSD 3-clause license is no longer being recognized.

Please see my repo here https://github.com/dyn4j/dyn4j and the LICENSE.md

### Expected behavior

The license is detected as BSD 3-clause

### Screenshots

N/A

### Additional context

This was working as of a week or two ago so it appears to be a regression.  I've tried to match the license content and format exactly (apart from name and years), but this did not seem to help.

mlinksva:
fixes #454

-----
[View rendered spec/fixtures/crlf-bsd/LICENSE.md](https://github.com/licensee/licensee/blob/universal_newline/spec/fixtures/crlf-bsd/LICENSE.md)

### DIFF
--- a/lib/licensee/project_files/project_file.rb
+++ b/lib/licensee/project_files/project_file.rb
@@ -40,6 +40,7 @@ module Licensee
         unless @content.valid_encoding?
           @content.encode!(ENCODING, **ENCODING_OPTIONS)
         end
+        @content.encode!(ENCODING, universal_newline: true)
 
         metadata = { name: metadata } if metadata.is_a? String
         @data = metadata || {}
--- a/spec/fixtures/crlf-bsd/LICENSE.md
+++ b/spec/fixtures/crlf-bsd/LICENSE.md
@@ -1,0 +1,27 @@
+Copyright (c) 2010-2020, William Bittle
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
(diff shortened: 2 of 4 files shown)


lib/licensee/project_files/project_file.rb
```diff
@@ -40,6 +40,7 @@ def initialize(content, metadata = {})
         unless @content.valid_encoding?
           @content.encode!(ENCODING, **ENCODING_OPTIONS)
         end
+        @content.encode!(ENCODING, universal_newline: true)
```
benbalter, line 43: Would it make sense to add `universal_newline: true` to the encoding options hash above and always run line 41 instead? Alternatively, I could see this living in content normalizer, replacing `\r\n?` with `\n`

mlinksva: I was not able to get adding `universal_newline: true` to the encoding options and always running line 41. Tests involving non-UTF-8-encoded license obtain `ArgumentError: invalid byte sequence in UTF-8`. I'm not sure why, but it seems the operations conflict with each other somehow.

I didn't include int he content normalizer because I'd read (but did not test) that string.encode is much faster than a regex, and it seemed fair to consider a pre-normalization step, having to do with the file type rather than the content.

However if you'd prefer it live in the content normalizer for consistency purposes, happy to move it there.

benbalter (approved):

PR #459, opened 2020-09-05
markdown normalization fixes

### LINKED ISSUE #434: Treat URLs to known licenses the same as full text license
### Is your feature request related to a problem? Please describe the problem you're trying to solve.

If a LICENSE file or similar entry consists solely of a link to a well known license URL (such as `http://www.apache.org/licenses/LICENSE-2.0.txt`), this is currently not detected as the "Apache-2.0" license.

Some dependencies declare their license intent by just a plain URL to the license (whether it's a proprietary license or a well known open source license). Currently a file with this content is detected as `None` or `NOASSERTION`.

### Describe the solution you'd like

A license described only by its URL, at least for well-known license URLs like this, should be interpreted as the associated license.  Such that `http://www.apache.org/licenses/LICENSE-2.0.txt` should be interpreted as `Apache-2.0` license.

### Describe alternatives you've considered

One workaround could be to detect if the contents of the file consist only of a URL, and then download that URL into a LICENSE file in order to match against it. But matching against a URL is much less text to sift through, compared to the full text of the license.

### Additional context

This particular project is using the [Google Open Source Licenses plugin for Gradle](https://developers.google.com/android/guides/opensource) in order to extract license information from each declared Gradle dependency. That content is then written to a seek-file with separate offset/length metadata. The contents can be extracted from the seek file by reading the start/length range from the file, for each dependency.  Some of those extracted licenses are full text (in some cases, the full Apache-2 license, or the Android SDK license, etc); but many of those are declared using only the license URL -- some of which are links to popular licenses like Apache-2.0, though some are considered proprietary licenses pointing to that company's terms of services.

Examples:
* Dagger (`com.google.dagger:dagger`) dependency extracts the full Apache-2.0 license text
* Kotlin Coroutines (`org.jetbrains.kotlinx:kotlinx-coroutines-core`) extracts only the URL: `http://www.apache.org/licenses/LICENSE-2.0.txt`
* Fabric plugin (`io.fabric.sdk.android:fabric`) extracts a URL to the company's terms: `https://fabric.io/terms`

The [kotlinx-coroutines POM](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-core/1.3.5/kotlinx-coroutines-core-1.3.5.pom) defines the license as:
```xml
  <licenses>
    <license>
      <name>The Apache Software License, Version 2.0</name>
      <url>https://www.apache.org/licenses/LICENSE-2.0.txt</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
```
But the Google `oss-licenses-plugin` extracts only the URL from this.

### LINKED ISSUE #435: Regression: Artistic License 2.0 (Markdown) no longer detected
### Describe the bug

As of [v9.11.0](https://github.com/licensee/licensee/releases/tag/v9.11.0) (Jan 2019), Licensee no longer detects [Artistic License 2.0](https://opensource.org/licenses/artistic-license-2.0) files (Markdown). They were detected correctly prior to [this commit](https://github.com/licensee/licensee/commit/d01660a751977271829339a3d57a181cbac76158).

1. Download an Artistic License 2.0 file in Markdown format, [e.g.](https://github.com/chocolateboy/wireless/blob/master/LICENSE.md)
2. Run the test script (see below)

#### Expected

```
ok: Artistic License 2.0
```

#### Actual

```
error: "Other"
```

## test.rb

```ruby
#!/usr/bin/ruby -W0

require 'licensee'

LICENSE = ARGV.shift || 'LICENSE.md'
WANT = 'Artistic License 2.0'

license = Licensee.license(LICENSE)&.name

if license == WANT
  puts "ok: #{license}"
else
  puts "error: #{license.inspect}"
end
```

mlinksva:
- headings were never matched unless at very beginning of string, change to beginning of line
- refactoring in d01660a751977271829339a3d57a181cbac76158 split out various bits of markdown, making ordering important; specifically for #435 alphanumeric bullets that are bolded would not be normalized fixed that by normalizing spans earlier

Fixes #435 

Changes the hash of odbl-1.0 because that license text natively includes markdown formatting

-----
[View rendered spec/fixtures/markdown-artistic/LICENSE.md](https://github.com/licensee/licensee/blob/remd/spec/fixtures/markdown-artistic/LICENSE.md)


benbalter (approved):

lcG392:
البحر غدار وله امان

PR #460, opened 2020-09-11
add mit-0 license

MarkAtwood:


MarkAtwood:
reviewing process, sorry

mlinksva:
You probably noticed already but in just in case not, there's an issue at https://github.com/github/choosealicense.com/issues/754

PR #465, opened 2020-10-14
Update release script

benbalter:
This updates the release script, such that `script/release --patch` would:

1. Bump the version number and commit the change
2. Tag the release
3. Push the tag and commit to GitHub
4. Push the release to rubygems.org

I've been using the command locally for release, but I'd like to move it to a script to make that process a bit more transparent.


mlinksva (approved):
Looks like gem-release does a lot of work.

PR #471, opened 2020-12-12
required_ruby_version = '>= 2.5'

mlinksva:
Rubocop suggests https://docs.rubocop.org/rubocop-performance/cops_performance.html#performancedeleteprefix which requires 2.5.

Not essential, but neither is supporting 2.4 which is EOL'd (happy to be wrong!)

Also try making CI test with 2.5, 2.6, 2.7.


benbalter (approved):
🚀

PR #477, opened 2021-01-23
Vendor SPDX XML license templates, use for more nuanced adjustment of expected license text variation

mlinksva:
Eventually want to create new matcher with these vendored templates, see https://github.com/licensee/licensee/issues/474#issuecomment-759219883 and below.

This PR starts by vendoring SPDX XML templates for licenses that licensee can detect, and does something marginally useful with those templates -- uses `<alt>` elements in those templates to more realistically account for expected variation in certain license texts than `[]` fields in the license texts vendored from choosealicense.com do.

This primarily impacts BSD licenses, which often have copyright holder names in the license text, rather than only in the copyright line, and refer to copyright holders or authors. The SPDX XML templates account for these, the choosealicense.com fields only the copyright line (and the latter shouldn't be modified to have more fields, that would only encourage more variation in the wild). I've added one fixture for a BSD-3-Clause license that falls just below detection threshold prior to this PR, but is detected with it.


lib/licensee/content_helper.rb
```diff
@@ -335,9 +335,21 @@ def fields_normalized_set
       @fields_normalized_set ||= fields_normalized.to_set
     end
 
-    def fields_adjusted_length_delta(other)
+    def spdx_alt_segments
+      @spdx_alt_segments ||= begin
+        path = File.expand_path "#{spdx_id}.xml", Licensee::License.spdx_dir
+        raw_xml = File.read(path, encoding: 'utf-8')
```
benbalter, line 341: To make things a bit more object-oriented, I'd expect to expose a `xml`-type method on the license itself, which content helper can call, rather than putting the logical to pull the license's XML representation into a helper shared between models.

lib/licensee/license.rb
```diff
@@ -64,6 +64,11 @@ def license_files
         @license_files ||= Dir.glob("#{license_dir}/*.txt")
       end
 
+      def spdx_dir
+        dir = ::File.dirname(__FILE__)
+        ::File.expand_path '../../vendor/license-list-XML/src', dir
```
benbalter, line 69:
```suggestion
        ::File.expand_path '../../vendor/license-list-XML/src', __dir__
```

benbalter (approved):
✨

lib/licensee/license.rb
mlinksva: Thanks! IIUC, the preceding line is now superfluous.

lib/licensee/content_helper.rb
mlinksva: Thanks. I don't want to expose anything yet but I moved the (private) method, you're right it doesn't make sense to have it shared.

PR #482, opened 2021-02-20
Simplify quote normalization

mlinksva:
Normalize all quote-like characters as single quotes.

Stop trying to preserve/normalize non-apostrophes as double quotes. This was imperfect (didn't work with quotes spread across multiple lines, and making it work across lines introduces other overly-greedy behavior.

For the dice matcher, resulting wordset sometimes has extra words (ending in s' that aren't apostrophes), sometimes has fewer words (words with quotes at beginning of multi-line quote). Overall no difference in matching quality, and no false negative for possessive, so take the simplification. And, the positives aren't really false given https://spdx.github.io/spdx-spec/appendix-II-license-matching-guidelines-and-templates/#5-punctuation -- arguably the wordset ought to include all punctuation.

Noticed opportunity for simplification when adding apostrophe normalization in https://github.com/licensee/licensee/pull/481


goneall:
As a reference - here's all the characters we normalize in the Java license compare tools: https://github.com/spdx/Spdx-Java-Library/blob/b9b04052d8d9776e5eea4290819243e95788503d/src/main/java/org/spdx/utility/compare/LicenseCompareHelper.java#L214

benbalter (approved):

PR #484, opened 2021-02-28
Rubocop auto-corrections

mlinksva:
Fix what can be fixed with rubocop auto-correct, and still have tests pass.


benbalter (approved):
Thanks for doing the needful.

PR #485, opened 2021-03-02
Content helper: Use simple delta for non-licenses

benbalter:
The content helper mixin is used in different objects. Licenses have a more advanced SPDX alt. segment-based delta (added via https://github.com/licensee/licensee/pull/477). Use that if it's present, otherwise, just return the simple delta. Specifically intended to prevent https://github.com/benbalter/coconductor from erroring out when `code_of_conduct.similarity` is called.


benbalter:
I confirmed in https://github.com/benbalter/coconductor/pull/26, that coconductor works on this branch.

mlinksva (approved):

PR #487, opened 2021-03-24
Suggestions under rubocop 1.12.0

mlinksva:
Mostly removing redundant begin blocks


benbalter (approved):
Ha. Thanks for tackling this. Ironically, those were added to appease line-length cops.

PR #506, opened 2021-09-29
Suggestions from newish rubocop and rubocop-spec versions

mlinksva:
I haven't tracked carefully but I think this means rubocop-1.2x and rubocop-rspec-2.5


benbalter (approved):
¯\\_(ツ)_/¯

Issue #526, opened 2022-01-19
How do i install?


slingring:
wheres the .exe i need to check a license on a github, but dont know where to find the license, and the github website said to use this so.

Tyler887:
In your Gemfile:
```ruby
gem 'licensee', git: 'https://github.com/licensee/licensee.git', branch: 'master'
```

slingring:
> In your Gemfile:
> 
> ```ruby
> gem 'licensee', git: 'https://github.com/licensee/licensee.git', branch: 'master'
> ```

The link says “Page Invalid”
And the hell is a gem file? I don’t use linux?

benbalter:
@slingring I've updated [the README](https://github.com/licensee/licensee#installation) with Ruby and Docker installation instructions. This project is not distributed as a compiled Windows binary, so you will need some familiarity with running either Ruby code or Docker containers within a windows environment. You could also consider using Codespaces if you do not want to set up the project locally.

PR #527, opened 2022-01-20
Add Dockerfile

benbalter:
I saw https://github.com/licensee/licensee/issues/526, and thought it would be easier to maintain a Dockerfile (and potentially down the line, a Docker image) than try to support Ruby environments on different platforms for users unfamiliar with Ruby.

This addition shouldn't affect day-to-day development, but allows users to run Licensee with the following:

1. `git clone https://github.com/licensee/licensee && cd licensee`
2. `docker build . --tag licensee`
3. `docker run licensee [COMMAND]`

It could also be used to facilitate local development, if people would prefer to develop within the docker container, rather than not their system directly.

I've also added a tiny CI job to build the docker container and verify it works to reduce maintenance and prevent regressions down the line.

-----
[View rendered docs/README.md](https://github.com/licensee/licensee/blob/docker/docs/README.md)

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -1,0 +1,21 @@
+name: Docker
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  build-image:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


mlinksva (approved):
Seems reasonable to me.

Hbutlercapone (commented):
Added Dockerfile #527

PR #528, opened 2022-01-21
Add package publisher

### LINKED ISSUE #526: How do i install?

Tyler887:
This should fix #526.

### DIFF
--- a/.github/workflows/gem-push.yml
+++ b/.github/workflows/gem-push.yml
@@ -1,0 +1,45 @@
+name: Ruby Gem
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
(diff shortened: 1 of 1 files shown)


mlinksva:
How would this fix #526? That's asking for a platform-specific executable. IIUC this would publish a gem based with each push to the default branch. There's already a gem for releases published to rubygems.

Tyler887:
> How would this fix #526? That's asking for a platform-specific executable. IIUC this would publish a gem based with each push to the default branch. There's already a gem for releases published to rubygems.

Thanks, I have withdrawed the request.

PR #532, opened 2022-02-08
docs: add a note about matcher confidences

villelahdenvuo:
Make the docs a bit more clear about matcher confidence.

-----
[View rendered docs/customizing.md](https://github.com/granodigital/licensee/blob/chore/docs/docs/customizing.md)

### DIFF
--- a/docs/customizing.md
+++ b/docs/customizing.md
@@ -22,6 +22,8 @@ project.license
 => #<Licensee::Licensee key="mit">
 ```
 
+**Note:** The package manager metadata matchers have a confidence of 90%, so you must set the confidence_threshold to 90 or lower for them to work.
+
 ### Matching project README license references
 
 Licensee supports the ability to take into account human readable references to licenses within the project's README, disabled by default. [There are reasons you may not want to use this](what-we-look-at.md). You can explicitly instruct licensee to take this information into account as follows:
@@ -31,3 +33,5 @@ project = Licensee.project("path/to/project", detect_readme: true)
 project.license
 => #<Licensee::Licensee key="mit">
 ```
+
+**Note:** The README license reference matchers have a confidence of 90%, so you must set the confidence_threshold to 90 or lower for them to work.


docs/customizing.md
```diff
@@ -22,6 +22,8 @@ project.license
 => #<Licensee::Licensee key="mit">
 ```
 
+**Note:** The package manager metadata matchers have a confidence of 90%, so you must set the confidence_threshold to 90 or lower for them to work.
```
mlinksva, line 25: I'm not sure the second part is true. If there is no license file and a package license id is found and package detection is turned on, I think licensee will return a conclusion of the license found in the license id. I believe the threshold currently is only pertinent to license texts. The hardcoded 90% is simply a signal to users of the command line tool. Need to test and look at the code to verify though.

mlinksva: > if I enable that setting and don't get any matches even if the license is specified in the package/readme file

Verified that is not the case:

```
% bundle exec licensee detect --packages https://github.com/mlinksva/pkg-license-only
License:        MIT
Matched files:  package.json
package.json:
  Confidence:  90.00%
  Matcher:     Licensee::Matchers::NpmBower
  License:     MIT
```
I think the right thing to do might be to remove the `Confidence: 90.00%` output in the case of a seeing a license indicated in a package file. We're confident that the package file has indicated the license in question. The confidence threshold in licensee currently only pertains to license text matching, not the overall license conclusion.

mlinksva:
Closing as this doc update does not match actual behavior, though it is understandably confusing. Suggest removing confidence output for package file matches instead https://github.com/licensee/licensee/pull/532#discussion_r806182512

docs/customizing.md
villelahdenvuo: @mlinksva However, if there is a LICENSE and a package file and the LICENSE file match is between 90-98% and the package file matches it's 90%, then licensee says there is no match. So it seems that package file is ignored completely if a LICENSE file exists.

I understand the reasoning, that the LICENSE file should be the main source of truth, so perhaps I can add a note to the docs saying that if a LICENSE file is found the package file match is ignored?

villelahdenvuo: So if one matcher returns `NOASSERTION` and another one returns a valid license, it will report all of them, but the final conclusion will be `NOASSERTION` as the sources don't agree. That makes sense indeed, sorry for the trouble explaining everything.

Perhaps I can update the docs at [what-we-look-at.md](https://github.com/licensee/licensee/blob/master/docs/what-we-look-at.md):

> If the project has multiple license matches (e.g. a package file match, a file named license or similar that matches the regular expressions) that don't only match one well known license (see below), Licensee won't return a license for the project, but all matches are returned in the licenses array.

And add a note with a link to https://github.com/licensee/licensee/blob/master/docs/usage.md#license-ruby-api for people who just open usage and don't read further. :)

mlinksva: Makes sense to me!

PR #533, opened 2022-02-13
Reset inverse confidence threshold when user override of confidence threshold is set

jonabc:
This relates to [changes in the licensed gem](https://github.com/github/licensed/pull/455) that allows users to customize Licensee's confidence threshold when evaluating OSS dependencies in an application.  I noticed that when changing the confidence threshold the inverse was not updating to match the change.  I made a small change here to reset the memoized value to nil when a user customizes the threshold value.


villelahdenvuo (approved):

benbalter (approved):
🚀

PR #536, opened 2022-02-17
docs: update note about multiple matches

villelahdenvuo:
I think we should add an example for getting all the license matches, but the current example only returns one license: `license = Licensee.license "/path/to/a/project"`

But I don't know how is the API for getting all the matches, is it just: `licenses = Licensee.licenses "/path/to/a/project"`?

-----
[View rendered docs/usage.md](https://github.com/granodigital/licensee/blob/chore/update-docs/docs/usage.md)
[View rendered docs/what-we-look-at.md](https://github.com/granodigital/licensee/blob/chore/update-docs/docs/what-we-look-at.md)

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -26,6 +26,8 @@ license.meta["permissions"]
 => ["commercial-use","modifications","distribution","private-use"]
 ```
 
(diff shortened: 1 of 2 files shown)


benbalter (approved):
🚀 Thanks for the docs improvements!

Issue #534, opened 2022-02-15
Doesn't choose license from standard LICENSE file if any other file with the word license appears, eg. check_license_exists.sh


HariSekhon:
### Describe the bug

When licensee detects two files with the word license in it, it gets confused and doesn't use the more obvious one, there is no priority for the standard `LICENSE` file so no license assertion is made.

### Steps to reproduce the behavior

```shell
$ git clone https://github.com/HariSekhon/DevOps-Bash-tools.git

$ cd DevOps-Bash-tools

$ licensee detect 
License:        NOASSERTION
Matched files:  LICENSE, check_license_exists.sh
LICENSE:
  Content hash:  4c2c763d64bbc7ef2e58b0ec6d06d90cee9755c9
  Attribution:   Copyright 2016 Hari Sekhon
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       MIT
check_license_exists.sh:
  Content hash:  0c66256c94dbbe7d947c7a9957bb79dd92960e0d
  License:       NOASSERTION
  Closest non-matching licenses:
    Unlicense similarity:     17.54%
    BSD-2-Clause similarity:  16.33%
    BSL-1.0 similarity:       14.29%
```

### Expected behavior

Expected it to determine the MIT license from the standard `LICENSE` file.

mlinksva:
Licensee intentionally matches files with additional words because sometimes actual license files have additional words in their names. I imagine we could add `.sh` to https://github.com/licensee/licensee/blob/06f5719b119513e00af73f4a578ed1ec0bc53d16/lib/licensee/project_files/license_file.rb#L17 to exclude shell scripts with license in their filenames though.

HariSekhon:
@mlinksva that ext regex list could start getting long as you may then need to also exclude `.py`, `.pl`, `.rb`, `.groovy` and any other languages which might have code with the word license in the filename...

mlinksva:
Ignoring random file with `license` in the name if a more standard named file exists in the project sounds reasonable. There's already some prioritizaiton of some filenames (see the linked file above) though it's just used as a way to organize the various regexes and doesn't factor into what files are looked at.

Implementing will require a bit of thought, leaving issue open.

HariSekhon:
[...]

PR #545, opened 2022-03-25
Add support for `.licensee.yml` ignored file pattern

benbalter:
This PR is a potential fix for https://github.com/licensee/licensee/issues/534.

It adds a new `.licensee.yml` file pattern. If Licensee is detecting a non-license file as a license, you can instruct licensee to ignore that file by adding a `.licensee.yml` file to the root of your project, with filenames to ignore under the `ignore` key. Ignore directives follow a similar glob pattern to `.gitignore` files, so you can also use wildcards to ignore multiple files at once.

To support the new `ConfigFile` concept, I created a `ContentFile` class to allow sharing of path, content, and encoding normalization logic between `ProjectFile` and `IgnoreFile`.

Thoughts on this approach?


lib/licensee/ignore_file.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module Licensee
+  class IgnoreFile < ContentFile
+    include Licensee::HashHelper
+    HASH_METHODS = %i[ignored_paths].freeze
+
+    FILENAME = '.licensee-ignore'
+
+    # Default IgnoreFile to maintain bacckward compatability
+    DEFAULTS = %w[xml go].map { |ext| "*.#{ext}" }.freeze
```
mlinksva, line 11: Should *.gemspec also be in this list? It's [excluded[ in OTHER_EXT_REGEX.

I see licensee has its own .licensee-ignore file, and licensee is probably why gemspec was excluded, so I guess it depends just how absolutely backwards compatible you might want to be (there are a variety of other gems with license in their names that could conceivably be affected, if one ran licensee on them).

lib/licensee/project_files/license_file.rb
```diff
@@ -12,8 +12,8 @@ class LicenseFile < Licensee::ProjectFiles::ProjectFile
       # Regex to match any extension except .spdx or .header
       LICENSE_EXT_REGEX = %r{\.(?!spdx|header)[^./]+\z}i.freeze
 
-      # Regex to match any extension except a few unlikely as license
-      # texts with complex filenames
+      # No longer used. Replaced by default IgnoreFile
+      # Maintianed for backward compatability
       OTHER_EXT_REGEX = %r{\.(?!xml|go|gemspec)[^./]+\z}i.freeze
```
mlinksva, line 17: To answer my own question looking at this, I guess this is needed for backwards compatibility because OTHER_EXT_REGEX is public thus part of the API.

lib/licensee/projects/fs_project.rb
```diff
@@ -34,7 +34,10 @@ def initialize(path, **args)
       def files
         @files ||= search_directories.flat_map do |dir|
           relative_dir = Pathname.new(dir).relative_path_from(dir_path).to_s
-          Dir.glob(::File.join(dir, @pattern).tr('\\', '/')).map do |file|
+
+          # Hidden files are excluded by default. Explicitly include IgnoreFile in file list
```
mlinksva, line 38: I don't think I understand what "Hidden files are excluded by default" means. Not recalling concept of a hidden file being used in licensee.

mlinksva (approved):
Seems like a good implementation if we're going to have this feature. A bit messy (additional special file) but I suppose there's demand for it.

One question: should it be possible to ignore a preferred file, eg `LICENSE`? I can see that could be useful say if someone has a primarily content project with `LICENSE` and `CONTENT-LICENSE` or something for code and content, and wants `CONTENT-LICENSE` only to be detected, but it also seems possibly misleading.

mlinksva:
Another quick/just a thought: conceivably other project-specific assertions for license detection might be useful in the future beyond an ignorelist. Perhaps the file should be `.licensee` which could initially only support an `ignore` section.

mlinksva:
Yet another quick thought: I don't know of any prior art/other license tooling with this feature, but if it exists would be good to know about.

lib/licensee/projects/fs_project.rb
benbalter: I found a better implementation that didn't require the comment. In short, Ruby's `Dir.glob` command ignores Unix-style hidden files (e.g., `.licensee.yml`), which you need to instruct it to include, via the `File::FNM_DOTMATCH` flag).

benbalter:
> Another quick/just a thought: conceivably other project-specific assertions for license detection might be useful in the future beyond an ignorelist. Perhaps the file should be `.licensee.yml` which could initially only support an `ignore` section.

I like that suggestion, and update the PR with that intention. Licensee will now read a `.licensee.yml` file, if present, which for now, supports the `ignore` directive, but could be used to support other config flags.

I backed out my `OTHER_EXT_REGEX` changes, preserving the behavior as is on `main` and removing the "default" ignorefile concept. My goal in this PR was to reduce ongoing maintenance and find tuning, and I felt like changing that regex behavior will have unforeseen and subtle consequences to how Licensee detects and prioritizes files, which I didn't want to get into at this time.

Thoughts with this approach?

seafoodmengthao (approved):

mlinksva:
@benbalter sorry for delay, ❤️ this approach, looks good to go to me.

mlinksva:
(Though I do still [wonder](https://github.com/licensee/licensee/pull/545#pullrequestreview-922344878) whether it ought to be possible to tell licensee to ignore a file named `LICENSE` or perhaps anything file currently [scored](https://github.com/licensee/licensee/blob/6b4871a3fc8854486f6af792cad5017cf69ae346/lib/licensee/project_files/license_file.rb#L35) above say `0.85`?)

Pandapip1:
Wouldn't it be easier for the file to directly specify which files use which licence?

SPDX could be used: https://spdx.dev/ids/

jsoref:
@mlinksva if this PR was rebased, what would it take to enable it to progress?

phuchau1989 (approved):

mlinksva:
Closing this after a few years 😬 I don't have bandwidth to take it back up and I'm just not sure yet another place to look for license metadata, which could be used disingenuously, is a net plus. And we have tweaked the regexes a few times since to include or exclude common license or non-license files, with no work needed by maintainers.

Issue #523, opened 2021-12-21
SIL OFL v1.1 not being detected


ghost:
### Description
I have got a project that uses the SIL OpenFont License v1.1, but it's detected as "Other".

### Steps to reproduce the behavior

1. Go to my project's GitHub page ~~here~~ (broken link, the project was deleted);
2. Open the `LICENSE` file;
3. See how it's copied from the [SIL OFL license's website](https://scripts.sil.org/cms/scripts/page.php?item_id=OFL_web);
4. See also how it does not present the permissions, limitations and conditions of the license, as when it's detected.

### Expected behavior
I expected GitHub to detect that the license is indeed SIL OFL v1.1, as it did with other projects, such as [JetBrains Mono](https://github.com/JetBrains/JetBrainsMono).

mlinksva:
Licensee currently doesn't know about reserved font names, and with one in the text it falls below its matching threshold.

I'd be open to a patch that accounts for reserved font names. It'd be kind of janky, but including in the copyright regex may be the right thing for now; in a way that'd reflect how reserved font names are part of the copyright text in https://github.com/spdx/license-list-XML/blob/23b581d6996c2db8b969d6c8529182ed1b27a628/src/OFL-1.1.xml#L14

mlinksva:
Fixed by #546

PR #546, opened 2022-04-07
Enable Exact matcher to detect OFL with a reserved font name

saltcandy123:
This pull request fixes issue #523.

The solution is to expand copyright regex to include the line that mentions a reserved font name, as mentioned on the issue.


mlinksva (approved):
Thanks! Works as suggested. Added an additional test to help me understand what's going on when I look in the future.

PR #589, opened 2022-09-21
Tighten license equality comparison

stanhu:
Previously if you attempted to compare `Licensee::License` to a boolean, it would fail with this obscure error:

```
NoMethodError Exception: undefined method `key' for true:TrueClass
```

This happened when Rails `ActiveSupport::Cache::Entry#dup_value!` attempted to duplicate the `License::License` by comparing it to boolean values.

We fix this by checking that `Licensee::License` is the other model.


mlinksva:
Thanks!

Unrelated to changes introduced here, but unclear to me why rubocop has one complaint, but _only_ under ruby 2.5:

> spec/licensee/license_spec.rb:3:1: C: Metrics/BlockLength: Block has too many lines. [424/420]

I'll look into this when I have a chance to run locally, but fixes welcome.

mlinksva (approved):

PR #594, opened 2022-10-05
Push image to Docker Hub

abitrolly:
Following #527 to push the image to Docker Hub.

This requires registering a username and setting up auth token at https://docs.docker.com/docker-hub/

Make it easier to launch `licensee` without downloading and building the image by hand.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -6,8 +6,11 @@ on:
   pull_request:
     branches: [master]
 
+env:
+  IMAGE_NAME: licensee/licensee
+
 jobs:
-  build-image:
+  build-and-push:
     runs-on: ubuntu-latest
 
     steps:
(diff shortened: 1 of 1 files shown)


mlinksva:
Think I understand the value, but not sure this is triggered by the right event. Currently it is run essentially as a smoke test, every time CI is run. Not sure it makes sense to push a new image in every such event. As part of release process I could see. Perhaps `-edge` or something on merge to master.

abitrolly:
It won't hurt to have an image to test latest changes. It is also possible to run the job on release too https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release and tag the image with `release.tag_name` in this case https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#release

alexsio27444 (approved):

webknjaz:
Alternatively, maybe publish to the GitHub Packages Registry that is bundled with this org? https://github.com/orgs/licensee/packages

abitrolly:
With DockerHub it is possible to do `docker run licensee` and with bundled repo it should be full path. Also, the GitHub quotas need to be checked against that DockerHub allows.

Malix-off:
This would be even more straightforward with GitHub package registry + rubygem integration

https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-rubygems-registry

mlinksva:
If someone really needs this to be published to the GitHub Packages Registry, feel free to shout or submit a PR.

Malix-Labs:
@mlinksva what is the point about closing this PR ?

PR #606, opened 2022-11-27
Remove licensee.gemspec contents from dump test to avoid dependency bump automatically causing failure

### LINKED ISSUE #599: Test JSON output for expected values, not exact file match so tests can pass after automated version bump
It's annoying that any dependency version bump requires running `script/dump-detect-json-fixture` and committing the results for tests to pass. This is caused by:

https://github.com/licensee/licensee/blob/16b2bf8e2fab709bec87d3ea84650e6b8c4410d0/spec/licensee/commands/detect_spec.rb#L72-L76

Suggest changing test to look for expected values rather than exact file match, delete `script/dump-detect-json-fixture` and `spec/fixtures/detect.json`.

mlinksva:
Fixes #599

### DIFF
--- a/spec/licensee/commands/detect_spec.rb
+++ b/spec/licensee/commands/detect_spec.rb
@@ -59,7 +59,7 @@ RSpec.describe 'detect command' do
 
   context 'json' do
     let(:arguments) { ['--json'] }
-    let(:expected) { JSON.parse fixture_contents('detect.json') }
+    let(:expected) { JSON.parse(fixture_contents('detect.json')).tap { |h| h['matched_files'][1].delete('content') } }
 
     it 'Returns a zero exit code' do
       expect(status.exitstatus).to be(0)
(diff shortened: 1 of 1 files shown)


spec/licensee/commands/detect_spec.rb
```diff
@@ -72,7 +72,7 @@
     it 'returns the expected output' do
       msg = +'`licensee detect --json` output did not match expectations. '
       msg << 'Run `script/dump-detect-json-fixture` and verify the output.'
-      expect(JSON.parse(stdout)).to eql(expected), msg
+      expect(JSON.parse(stdout).tap { |h| h['matched_files'][1].delete('content') }).to eql(expected), msg
```
mlinksva, line 75: Yes the same `.tap` call again. This isn't required anywhere else so duplication strikes me as better than a helper method elsewhere. Happy to be wrong!

PR #625, opened 2022-12-30
match https://spdx.org/licenses/equivalentwords.txt more precisely

mlinksva:
It could make sense to vendor list from https://github.com/spdx/license-list-XML/blob/main/equivalentwords.txt but sub-license on both sides would require some handling.


benbalter (approved):

BLUEOO73 (approved):
1inç cuzdanlma aktar Nurcan Tuncay

PR #682, opened 2023-11-13
Try grouping actions updates

mlinksva:
So as to not get several PRs following https://github.com/licensee/licensee/pull/670#issuecomment-1785619639


benbalter (approved):
Nice! Good call.

PR #684, opened 2023-11-17
License-path command fix

### LINKED ISSUE #667: license-path command return wrong path.
### Describe the bug

The `license-path` command return the wrong path to the license.

1. For example, run `licensee license-path /path/to/my/repo`.
2. It will print the current working directory as in: `/current/working/directory/LICENSE`.

### Expected behavior

It should print `/path/to/my/repo/LICENSE` or just plain `LICENSE`.

boimw:
Fixes #667

Steps to validate:
1. run `licensee license-path /path/to/my/repo` that is not current repo
2. validate that output is correct


mlinksva (approved):
Thanks @boimw for this fix (and patience on the review)!

PR #694, opened 2023-12-18
Remove support for Ruby 2.7

mlinksva:
EOL since 2023-03-31 https://www.ruby-lang.org/en/downloads/branches/

Various rubocop suggestions for Ruby 3.0+


dmoss18:
Just FYI this is a breaking change and you included it in a patch release. Anybody still on ruby 2.7 who goes from 9.16.0->9.16.1 can no longer bundle install. Should this have been a major release instead of a patch?

mlinksva:
`bundle install` simply not installing 9.16.1 is the reason I considered it OK to make it a patch release based on what is in the release, just a small bugfix and various version updates. Nothing will break, you just don't get the new version. And if you try to manually set the new version, you still won't get it.

If there's a Ruby (or broader) consensus to always make required language version (or say changes in platform support) mandate a major version change, happy to take this onboard for future releases.

Issue #664, opened 2023-09-12
A function to obtain different licenses for each Github tag


yongyongsi:
### Is your feature request related to a problem? Please describe the problem you're trying to solve.

It appears that in some cases, the license may vary for each tag or commit hash.
this Github repository example provided below, this situation is evident.

https://github.com/platonai/PulsarRPA/tree/0.5.0
https://github.com/platonai/PulsarRPA

 I'm currently using Licensee in a Docker environment through a CI command, but even when I provide different arguments for each commit hash, the license for the 'master' branch is consistently displayed.

### Describe the solution you'd like

I wish it were possible to obtain the license associated with a specific GitHub tag when provided with that tag.

### Describe alternatives you've considered

It would be great if we could retrieve the license using a link that includes a tag or commit hash or by providing the tag as a separate argument.

mlinksva:
> I'm currently using Licensee in a Docker environment through a CI command, but even when I provide different arguments for each commit hash, the license for the 'master' branch is consistently displayed.

How are you providing those arguments? For a remote repo? Or one that you've locally cloned? Can you give an example of the invocation?

yongyongsi:
1. Clone this project to my local pc
`git clone https://github.com/licensee/licensee && cd licensee`

2. docker build
`docker build . --tag licensee

3. execute command
`docker run licensee detect {owner/repo} --remote

In this case, when I included a link with the version to obstain a specific version of the license, it provided the license for the 'master' version.

`docker run licensee detect https://github.com/platonai/PulsarRPA/blob/v1.7.0/ --remote`
When i executed the command for version 1.7.0 of this project. it displayed the AGPL license from the 'master' brance, even though vesion 1.7.0 is supposed to have the Apache License.

`docker run licensee detect https://github.com/platonai/PulsarRPA/blob/0.5.0/ --remote`
Even for version 0.5.0, the license information is present in the 'COPYING' file, but it is incorrectly marked as the AGPL License in the 'LICENSE' file.

So, I cloned both of these projects locally and then ran 'licensee', confirming that the licenses for each project are displayed correctly.
What I want is for the licenses to be correctly displayed with just the version information provided in the link, without the need to clone the projects separately.

mlinksva:
Thanks for explaining. Currently licensee extracts `owner/repo` and ignores anything trailing:

[...]

PR #722, opened 2024-03-10
Implement ability to fetch license for ref via github

jarrydlee:
There was some discussion about adding the ability to fetch the license at a given tag in https://github.com/licensee/licensee/issues/664. I want to make it a little more generic and allow for any ref that github supports, so I added the `--ref` flag to accomplish this. 

Partial ref formats are supported, ie `main` instead of `refs/heads/main`, and although I couldn't find the exact resolution github uses, I can only imagine it uses the default specified [here](https://mirrors.edge.kernel.org/pub/software/scm/git/docs/gitrevisions.html).

I had thought about building this into the `GitProject` too, but didn't want to potentially break the `commit` option that already exists.

Small disclaimer, I do not write ruby, so please make sure things make sense.
cc @benbalter


benbalter (approved):
Thanks for this!

jarrydlee:
No worries! What does the release process look like from here, and is there anything you need from me?

spec/licensee/projects/git_hub_project_spec.rb
```diff
@@ -109,4 +109,20 @@
       expect { subject.license }.to raise_error(described_class::RepoNotFound)
     end
   end
+
```
mlinksva, line 112: Do the tests below test that licensee can successfully detect a license on the ref provided, and not eg a different license on the default branch? Not certain needed, just seems an obvious question.

mlinksva (changes requested):
https://github.com/licensee/licensee/blob/main/docs/command-line-usage.md should be updated along with this.

spec/licensee/projects/git_hub_project_spec.rb
jarrydlee: I thought about the helpfulness of writing that test since most of the test will be mocked behavior of github APIs. 
Although I think it would be helpful to add the test to assert the request we're stubbing correctly contains the query params. I'll add this in along with the cli doc changes 👍

docs/command-line-usage.md
```diff
@@ -60,6 +60,7 @@ Options:
                                  # Default: 98
   [--license=LICENSE]            # The SPDX ID or key of the license to compare (implies --diff)
   [--diff], [--no-diff]          # Compare the license to the closest match
+  [--ref=REF]                    # The name of the commit/branch/tag to search
```
mlinksva, line 63: Should this and the detect.rb option specify github.com since it'll only work with repos there with this implementation? IIUC the option will silently not have any effect if executed against a local repo, for instance, right?

mlinksva (changes requested):
Added one tiny documentation change to consider.

Thanks for adding more mocked tests. I'll test personally and merge then cut a release if that goes well as I have time in the next few days.

docs/command-line-usage.md
jarrydlee: Ah yeah, I'd just pulled the default `help detect` output. I added a note that ref only works for github PATHs.

lib/licensee/commands/detect.rb
```diff
@@ -13,6 +13,7 @@ class LicenseeCLI < Thor
   option :confidence, type: :numeric, default: Licensee.confidence_threshold, desc: 'Confidence threshold'
   option :license, type: :string, desc: 'The SPDX ID or key of the license to compare (implies --diff)'
   option :diff, type: :boolean, desc: 'Compare the license to the closest match'
+  option :ref, type: :string, desc: 'The name of the commit/branch/tag to search (github.com PATHs only)'
```
mlinksva, line 16: `PATHs` is somewhat confusing terminology, wonder if this should just be:
```suggestion
  option :ref, type: :string, desc: 'The name of the commit/branch/tag to search (github.com only)'
```

I might be missing something but a chatbot seems to agree 🙃 

> Certainly! Let's dive into the world of Git and GitHub paths.
> 
> 1. **Git Reference (Git Ref)**:
>     - A **Git reference** (often called a **git ref**) is a file that contains a **Git commit SHA-1 hash**.
>     - When referring to a specific Git commit, you can use the **Git reference** as an easy-to-remember name instead of dealing with the raw hash.
>     - Git references can be rewritten to point to new commits. For example:
>         - A **branch** is a Git reference that stores the latest Git commit hash for that branch.
>         - A **tag** is another type of Git reference that points to a specific commit (usually used for releases).
>     - Instead of saying "commit `a3b7f9c2e`," you can say "the `main` branch" or "the `v1.0` tag."
> 
> 2. **GitHub PATH**:
>     - The term "GitHub PATH" might be a bit confusing because it's not a standard Git concept.
>     - However, I suspect you might be referring to **GitHub REST API endpoints** related to Git references.
>     - These endpoints allow you to interact with references (branches, tags, etc.) in your Git database on GitHub.
>     - For example, you can use the GitHub API to:
>         - List matching references based on a name (e.g., branches or tags).
>         - Retrieve information about a specific reference (e.g., commit SHA-1 hash).
>         - Create or update references.
>     - The **GitHub PATH** here refers to the URL path you use to access these API endpoints.
> 
> 3. **Example API Endpoint**:
>     - To list matching references (branches or tags) in a GitHub repository, you can use the following API endpoint:
>         - `GET /repos/{owner}/{repo}/git/matching-refs/{ref}`
>     - Here:
>         - `{owner}` is the account owner (case-insensitive).
>         - `{repo}` is the repository name (without the `.git` extension).
>         - `{ref}` can be a commit SHA, branch name (e.g., `heads/main`), or tag name (e.g., `tags/v1.0`).
>     - The response will include information about the matching references.
> 
> 4. **Example Usage**:
>     - Suppose you want to list matching references for the `main` branch in a repository owned by `octocat`. You'd make a request like this:
>         - `GET /repos/octocat/my-repo/git/matching-refs/heads/main`
> 
> Remember, Git references and GitHub API endpoints play a crucial role in managing version control and collaboration within repositories! [🌟](https://docs.github.com/en/rest/git/refs)¹²³..
> 
> Source: Conversation with Bing, 3/[1](https://docs.github.com/en/rest/git/refs)[2](https://git-scm.com/book/en/v2/Git-Internals-The-Refspec)/2024
> (1) REST API endpoints for Git references - GitHub Docs. https://docs.github.com/en/rest/git/refs.
> (2) Git - The Refspec. https://git-scm.com/book/en/v2/Git-Internals-The-Refspec.
> ([3](https://superuser.com/questions/614978/how-to-reference-a-git-repository)) path - How to reference a git repository? - Super User. https://superuser.com/questions/614978/how-to-reference-a-git-repository.
> (4) Git - git-show-ref Documentation. https://git-scm.com/docs/git-show-ref.
> (5) undefined. https://api.github.com/repos/OWNER/REPO/git/matching-refs/REF.
> (6) undefined. https://github.com/schacon/simplegit-progit.

(First response related to file and executable paths, which I guess will be what most people associate path with.)

mlinksva (changes requested):

lib/licensee/commands/detect.rb
mlinksva: Alternately, 
```suggestion
  option :ref, type: :string, desc: 'The name of the commit/branch/tag to search (github.com refs only)'
```

jarrydlee: The reason I chose `PATH` is because it's referring to the `licensee detect [PATH]` in the same doc message. That said, I'm happy to commit one of these, I'd prefer the first.

mlinksva:
Great, it works for me. Picked an example from https://en.wikipedia.org/wiki/Software_relicensing

```
$ bundle exec licensee detect https://github.com/NatronGitHub/Natron
License:        NOASSERTION
Matched files:  LICENSE.txt, LICENSE_SHORT.txt
LICENSE.txt:
  Content hash:  32108116603c30687d8d0d2f77f140fb6ecea082
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       GPL-2.0
LICENSE_SHORT.txt:
  Content hash:  f00bd04044d10c152bdcd2429db71f10020495ce
  License:       NOASSERTION
  Closest non-matching licenses:
    Zlib similarity:        34.78%
    MIT-0 similarity:       31.87%
    PostgreSQL similarity:  30.27%
$ bundle exec licensee detect --ref=53e14dc https://github.com/NatronGitHub/Natron
License:        MPL-2.0
Matched files:  LICENSE.txt
LICENSE.txt:
  Content hash:  820048a1dbef5dfac65547bd9eb935beb76b2257
  Confidence:    100.00%
  Matcher:       Licensee::Matchers::Exact
  License:       MPL-2.0
$ bundle exec licensee detect --ref=ac4965e https://github.com/NatronGitHub/Natron
License:  None
```

mlinksva (approved):

PR #795, opened 2024-11-22
Prioritize files likely to have actual license content

mlinksva:
Lower "filename scoring" of files like `copyright` so that most relevant files are reported on first.


ashleywolf (commented):
LGTM

PR #820, opened 2025-02-21
Add support for REUSE-style `LICENSES/` directories

### LINKED ISSUE #737: Scan `LICENSES` folder (and maybe other common folders)
### Is your feature request related to a problem? Please describe the problem you're trying to solve.

If you want to have multiple license files in your project, you may want to add them inside a subfolder.

The [REUSE](https://reuse.software) project which aims to make licensing more strict require the licenses files to be in a `LICENSES` folder in the project root folder.

I would like to follow REUSE recommendations and be able to run the associated linter, but if I put my licenses in a folder they are no longer recognized and displayed by GitHub which kinda defeat the point of having clear licensing.

### Describe the solution you'd like

Scan not only the root folder but also the `LICENSES` folder for licenses (and maybe other common folder names?).

Would be solved by https://github.com/licensee/licensee/issues/603, but this would have a bigger performance hit I guess.

### Describe alternatives you've considered

None

### Additional context

None

erickt:
[reuse-tool] is becoming a popular way to manage licenses, and is now used by projects like [Rust] and [curl]. License files must be placed in a [LICENSES/] directory. This patch simply teaches the `licensee` tool to also search in this directory for licenses.

Along the way, since this changes the output to include the directory by implementing an old TODO to get `Project_file.path` to be the absolute path to the project root, migrates `detect.rb` over to using it.

Closes #737

[reuse-tool]: https://reuse.software/
[Rust]: https://github.com/rust-lang/rust/tree/master/LICENSES
[curl]: https://github.com/curl/curl/tree/master/LICENSES
[LICENSES/]: https://reuse.software/spec-3.3/#license-files


lib/licensee/project_files/project_file.rb
```diff
@@ -44,13 +44,13 @@ def initialize(content, metadata = {})
         @data = metadata || {}
       end
 
-      # TODO: In the next major release, filename should be the basename
-      # and path should be either the absolute path or the relative path to
-      # the project root, but maintaining the alias for backward compatability
       def filename
```
erickt, line 47: I'm not sure how this is used, so this may be a breaking change.

lib/licensee/projects/git_project.rb
```diff
@@ -68,7 +68,16 @@ def load_file(file)
       #  :name - the file's path relative to the repo root
       #  :oid  - the file's OID
```
erickt, line 69: I should document here and elsewhere that we're also including `:dir` in the hash.

lib/licensee/projects/project.rb
```diff
@@ -8,16 +8,18 @@
 module Licensee
   module Projects
     class Project
-      attr_reader :detect_readme, :detect_packages
+      attr_reader :detect_readme, :detect_packages, :detect_reuse_licenses_dir
```
erickt, line 11: Should this be an optional feature, or always on?

erickt (commented):
This also needs tests for filesystem and plain old git projects.

mlinksva (changes requested):
Thanks for this, a few comments/requests:

* I'd call/invoke this as `licenses_dir` rather than `reuse_licenses_dir` as I think we'd just be looking for license files in the LICENSES dir rather than attempting to follow the REUSE spec (especially across versions of it), and if there were a file in that dir not named according to the spec but likely a license anyway, it probably makes sense to detect
* But to be useful I imagine we've got to take into account the spec a bit, which doesn't IIUC include `LICENSE` or similar in the filenames in `LICENSES` rather `{SPDX_ID}.txt`;  these should be detected and tested for
* I'd prefer the tests to be `fixtures` based rather than webmock, easier to understand and less fragile.

lib/licensee/projects/git_project.rb
erickt: Done in the latest version.

lib/licensee/projects/project.rb
erickt: Removed in the latest version, but I can restore if you want this to be an option.

erickt:
Thanks @mlinksva!

> I'd call/invoke this as licenses_dir rather than reuse_licenses_dir as I think we'd just be looking for license files in the LICENSES dir rather than attempting to follow the REUSE spec (especially across versions of it), and if there were a file in that dir not named according to the spec but likely a license anyway, it probably makes sense to detect

I ended up removing the `licenses_dir` since it seemed to make sense to just implement this functionality in the `licensed_file.rb` matcher. Please let me know if you want this to be optional.

> But to be useful I imagine we've got to take into account the spec a bit, which doesn't IIUC include LICENSE or similar in the filenames in LICENSES rather {SPDX_ID}.txt; these should be detected and tested for

I think I got matching for the SPDX_ID correct. I couldn't find a formal regex for them, so I copied what you did with the SPDX matcher.

> I'd prefer the tests to be fixtures based rather than webmock, easier to understand and less fragile.

Fixed and added some fixture tests.

lib/licensee/project_files/license_file.rb
```diff
@@ -64,6 +64,8 @@ class LicenseFile < Licensee::ProjectFiles::ProjectFile
         ^(creative\ commons\ )?Attribution-(NonCommercial|NoDerivatives)
       /xi
 
+      SPDX_REGEX = /\A[a-z\-0-9. +()]+#{PREFERRED_EXT_REGEX}+/i
```
mlinksva, line 67: I suspect this would fail to match some LicenseRef-[idstring] files because idstring need not start with a number https://reuse.software/spec-3.3/#license-files

lib/licensee/projects/git_project.rb
mlinksva: This looks like a potentially significant API change. Is it necessary?

spec/fixtures/licenses-dir-with-multiple-license-files/LICENSES/MIT.txt
```diff
@@ -0,0 +1,21 @@
+MIT License
```
mlinksva, line 1: Are there corresponding tests for these fixtues/

spec/fixtures/webmock-licenses-dir-with-multiple-license-files/dir.json
```diff
@@ -0,0 +1,26 @@
+[
```
mlinksva, line 1: I was thinking we'd not have any webmock tests for this feature.

mlinksva (changes requested):

spec/fixtures/webmock-licenses-dir-with-multiple-license-files/dir.json
erickt: Ah ok, I get you now. I'll instead inline these json files into `git_hub_project_spec.rb` in my next revision.

spec/fixtures/licenses-dir-with-multiple-license-files/LICENSES/MIT.txt
erickt: I think so, I added this to the `spec/fixtures/fixtures.yml`, which I believe is tested by `spec/fixture_spec.rb`. It's a fairly basic test though. Is there another test that lists out all the found licenses, or should I extend `spec/fixture_spec.rb` to support it?

lib/licensee/projects/git_project.rb
erickt: Which change are you referring to? If it is that we're including `:dir` in the hash, as far as I could tell all 3 backends already included `:dir`, it just wasn't documented.

If it's that I changed `files` to also include the `LICENSES/` files, yes this does change things. Would you prefer I could instead add a new API to fetch files from a subdirectory?

PR #907, opened 2025-11-21
Add cooldown settings for package ecosystems

mlinksva:
Follow advice of https://blog.yossarian.net/2025/11/21/We-should-all-be-using-dependency-cooldowns


benbalter (approved):
TIL'd

PR #914, opened 2025-12-16
Add Ruby 4.0 to CI matrix

mlinksva:
This may not work until 4.0 is released, might depend on how setup-ruby resolves (does it use previews not expressly stated)?


mlinksva:
Nice, setup-ruby does pull in 4.0.0-preview2.

PR #996, opened 2026-05-01
Make rugged an optional dependency

### LINKED ISSUE #66: Make dependency on rugged optional
Now that there is support for filesystem-based projects, how about making the dependency on rugged optional, and only use it if it's already installed? The main reason for this being that rugged, as a native extension, can be a pain to install on some platforms, and after all a Git working tree can also be scanned by the filesystem scanner.

stanhu:
rugged is a native extension that requires cmake and openssl to build, which can be painful on some platforms. Since filesystem-based scanning covers the common case, move rugged to a development dependency and fall back to FSProject transparently when it is not installed.

Users who scan bare Git repositories or repos without a working tree can opt in by adding rugged to their Gemfile or running gem install rugged.

Fixes #66

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -39,6 +39,22 @@ On Windows, the last line needs to include the Ruby interpreter:
 
     bundle exec ruby bin\licensee
 
+### Git repository scanning (optional)
+
+By default, licensee scans projects using the filesystem. If you want to scan bare Git repositories or repositories without a working tree (e.g., on a Git server), install the [rugged](https://github.com/libgit2/rugged) gem, which provides Ruby bindings for libgit2:
+
+    gem install rugged
+
+Or add it to your `Gemfile` alongside licensee:
+
+```ruby
+gem 'rugged'
+```
+
+When rugged is available, `Licensee.project` will use it automatically. If rugged is not installed, licensee falls back to filesystem-based scanning transparently.
+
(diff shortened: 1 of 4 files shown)


stanhu:
@benbalter This is follow-up of the attempt in https://github.com/licensee/licensee/pull/78. What do you think?

mlinksva (changes requested):
Seems reasonable to me. Because it's a behavior change it should probably go in a new major version. I will likely cut a final 9.x version before merging this.

Is there a reasonable test or two that could be added for this new behavior?

stanhu:
> Is there a reasonable test or two that could be added for this new behavior?

I added a unit test. We could add a test that omits `rugged` entirely, but we'd need to drop the dependency from the `.gemspec` and out into a bundler group in `Gemfile`.

mlinksva (approved):

mlinksva:
Thanks. I intend to merge this and cut a 10.0.0 release after first cutting a new 9.x release. It might be a few days.

mlinksva:
10.0.0 released

stanhu:
@mlinksva Thank you so much!

sschuberth:
> @mlinksva Thank you so much!

Thank _you_ so much, @stanhu, for fixing this long-standing issue!

PR #1023, opened 2026-07-07
Add a matcher for pyproject.toml

v--:
The pyproject.toml file has been standardized in 2020 (in [PEP 621]), so it is now common to all package managers.

Although Python projects tend to have LICENSE files, I felt that adding an auxiliary detection mechanism would do no harm.

[PEP 621]: https://peps.python.org/pep-0621/


mlinksva:
@v-- thanks. The copilot review comments look plausible to me, but I'm unfamiliar with pyproject.toml. Please take a look.

v--:
@mlinksva GitHub decided to silently ignore the comment I wrote above, so I rewrote it here. It was intended for you anyway.

1. The `license = { text = "<name>" }` format suggested in the review has been deprecated in PEP 639 ([link to deprecation](https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys)). 
2. TOML may be a simple language, but it is not regular, and parsing TOML tables properly requires a library. If licensee (or ruby's stdlib) decides to include one, my code (as well as the Cargo.toml code) can be refactored.
3. There are other format variations not covered by a simple field check, but I believe most are outside the scope of this project.

PS: I (force-)pushed a new commit with a erroneous quote removed.

mlinksva (approved):

